PR #1, opened 2021-06-01
pip: Properly lookup source version

Joshua-Anderson:
Fixes errors when trying to install pip package with `pip install -e`


LeStarch (approved):
We need to make the same change to fprime-gds too

PR #3, opened 2021-06-03
Verify target exists before building

Joshua-Anderson:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Before attempting to build a target, verify that the target actually exists.

## Rationale

There are some cases where local targets don't exist in a particular directory (Try `fprime-util build` in the root of the fprime repo for example`). Right now, this isn't caught and fprime-util prints a cmake usage message with no error details.

Target lookup now looks like this:

- Lookup target in list of all possible targets
- Use target to determine build type to load
- Load build
- Check if local target actually exists within loaded build

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -360,6 +360,14 @@ class Build:
             "auto_location": auto_location,
         }
 
+    def check_target(self, cwd: Path, target: Target):
+        if isinstance(target, GlobalTarget):
+            return
+
+        valid_cmake_targets = self.cmake.get_available_targets(self.build_dir, cwd)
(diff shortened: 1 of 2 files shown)


Joshua-Anderson:
Closing in favor of #4.

It turns out that `fprime-util build` in the root of fprime was a special case triggered by newer versions of cmake erroring when given an empty target argument. fprime-util already has special handling for the case where a target was missing where fprime-util would regenerate to pick up potential new targets. The error messages aren't quite as good as this change would print (Ex. `make: *** No rule to make target 'Svc'.  Stop.`), but I think they're good enough and not worth the complexity of having this change attempt to automatically regenerate.

PR #5, opened 2021-06-09
Cookiecutter

aidan-wagner:
| | |
|:---|:---|
|**_Affected Component_**| fprime-util new |
|**_Related Issue(s)_**|  |

---
## Change Description

Allows the user the option to set the cookiecutter template. By default, a local template is used.

## Rationale

This gives the users flexibility in using a cookiecutter of their choice

To test this feature, you can run:

```
fprime-util new
```

with a settings.ini file and without a settings.ini file.
## Future Work

In the future, the local cookiecutter will be refined and improved to work better with the recent changes to F'.


src/fprime/cookiecutter_templates/cookiecutter-fprime-component/CHANGELOG.rst
```diff
@@ -0,0 +1,9 @@
+Changelog
```
Joshua-Anderson: Do we need this changelog file?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/README.rst
```diff
@@ -0,0 +1,177 @@
+===============================
```
Joshua-Anderson: Do we need this readme?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
Joshua-Anderson: fprime-tools is apache2 licensed. Dropping in different a license within the code is unexpected and will probably lead to confusion. We should reach out to Sterling about how to properly credit him while possibly moving this code under the apache 2.0 license.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
```diff
@@ -0,0 +1,88 @@
...
+import datetime
+import os
+from os.path import join
+from fprime.fbuild.settings import IniSettings
+from fprime.fbuild.builder import Build
+from fprime.fbuild.interaction import make_namespace
+from pathlib import Path
+
+try:
+    from click.termui import secho
```
Joshua-Anderson: We should either require click.termui for fprime-tools or not require it. I don't think there's any benefit to adding the complexity of conditionally supporting it.

It seems like you're just using click for ANSI color printing warning messages, is there any reason to not just write your own warn function that adds the ASCI escape sequences you're looking for?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
```diff
@@ -0,0 +1,88 @@
...
+            secho(line, fg="white", bg="red", bold=True)
+
+
+def replace_contents(filename, what, replacement, count = 1):
+    with open(filename) as fh:
+        changelog = fh.read()
+    with open(filename, 'w') as fh:
+        fh.write(changelog.replace(what, replacement, count))
+
+if __name__ == "__main__":
```
Joshua-Anderson: To follow python best practices, I'd recommend reorganizing all code following from here into a main function, then calling the main function from the main if statement at the end of the file.


```python
if __name__ == "__main__":
    main()
```

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
```diff
@@ -0,0 +1,88 @@
...
+
+################################################################################
+
+    Now you can edit your {{ cookiecutter.component_name }}Ai.xml file
+    define the component to your liking
+
+    In addition, a sdd.md file has been created in the docs directory 
+    for you to document your component
+
+    If the project root was not found, you will need to add this component to
```
Joshua-Anderson: Can we update this usage message to not print this missing project root message if the project root was fine and warn about the missing project root and what to do about it if the project root is missing?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
```diff
@@ -0,0 +1,88 @@
...
+    path_to_fprime = IniSettings.find_fprime(cwd)
+
+
+with open("{{ cookiecutter.component_name }}ComponentAi.xml", "r") as s:
+    lines = s.readlines()
+    s.close()
+
+#Write schema location and namespace to .xml file
+with open("{{ cookiecutter.component_name }}ComponentAi.xml", "w") as s:
+    lines.insert(1, '<?xml-model href="' + str(path_to_fprime) + '/Autocoders/Python/schema/default/component_schema.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>\n')
```
Joshua-Anderson: We no longer use an XML schema in our F' xml files, so we can remove the code that injects this.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/.cookiecutterrc
```diff
@@ -0,0 +1,20 @@
+# This file exists so you can easily regenerate your project.
```
Joshua-Anderson: Is there any case where we'd want to regenerate an existing F' component? I'm not sure we need support for cookiepatcher.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/.gitignore
```diff
@@ -0,0 +1,2 @@
+# ignore backup files
```
Joshua-Anderson: Do we need a per-component .gitignore? I think most projects use a single gitignore at the top level of their project

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
...
+#     Add your dependencies here
+# )
+
+register_fprime_module()
+
+set(UT_SOURCE_FILES
+  "${CMAKE_CURRENT_LIST_DIR}/test/ut/main.cpp"
+  "${CMAKE_CURRENT_LIST_DIR}/test/ut/Tester.cpp"
+  "${CMAKE_CURRENT_LIST_DIR}/test/ut/TesterBase.cpp"
+  "${CMAKE_CURRENT_LIST_DIR}/test/ut/GTestBase.cpp"
```
Joshua-Anderson: @LeStarch I'm not sure the full history of this, but If I recall correctly the autocoder now always generates the `TesterBase` and `GTestBase` files with the rest of the autocode in addition to generating them when someone runs `impl --ut`. I'm pretty sure we don't need to add these files to cmake and we may want to make sure the autocoder no longer generates them with `impl --ut`

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/.gitignore
```diff
@@ -0,0 +1 @@
+*.html
```
Joshua-Anderson: I don't think there's anything that would generate HTML in the component docs directory, so I'm not sure this gitignore is needed.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/{{cookiecutter.component_name}}.md
```diff
@@ -0,0 +1,22 @@
+# {{cookiecutter.component_name}} Component Dictionary
```
Joshua-Anderson: @LeStarch Do we need a manual component dictionary? Isn't this the point of the autocoded html dictionary? This seems difficult to keep up to date.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
```diff
@@ -0,0 +1,85 @@
...
+            <comment>
+            The ping output port
+            </comment>
+        </port>
+    </ports>
+{% endif %}
+{% if cookiecutter.commands == "yes" %}
+    <commands>
+{%- if cookiecutter.component_kind == 'passive' %}
+        <command kind="sync" opcode="0" mnemonic="PR_StopPings">
```
Joshua-Anderson: Some of our example components have these commands to disable the ping response, but that's mostly for demo purposes.

Did you add this so there would be an example command in the newly generated component? If that's the case maybe it's worth making it more generic and calling it something like "ExampleCommand"

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
```diff
@@ -0,0 +1,85 @@
...
+{%- endif %}
+            <comment>
+            Command to disable ping response
+            </comment>
+        </command>
+    </commands>
+{% endif %}
+{% if cookiecutter.parameters == "yes" %}
+    <parameters>
+        <parameter id="5" set_opcode="6" save_opcode="1" name="parameter1" data_type="U32" default="10">
```
Joshua-Anderson: I like the idea of having an example parameter, but maybe only add one instead of two?

I'm not sure familiar with parameters, do you know why you picked ID 5 and set_optcode 6? Is it not possible to set them to 1?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
```diff
@@ -0,0 +1,85 @@
...
+        <parameter id="7" set_opcode="8" save_opcode="3" name="parameter2" data_type="I16" default="11">
+            <comment>
+            A test parameter
+            </comment>
+        </parameter>
+    </parameters>
+{% endif %}
+{% if cookiecutter.telemetry == "yes" %}
+    <telemetry>
+        <channel id="0" name="PR_NumPings" data_type="U32">
```
Joshua-Anderson: Similar to what I said about the stop ping commands above, what do you think about calling this something more generic, like `TestChannel`?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
```diff
@@ -0,0 +1,85 @@
...
+            <comment>
+            Number of pings received
+            </comment>
+        </channel>
+    </telemetry>
+{% endif %}
+{% if cookiecutter.events == "yes" %}
+    <events>
+        <event id="0" name="PR_PingsDisabled" severity="ACTIVITY_HI" format_string = "PingReceiver ping responses disabled" >
```
Joshua-Anderson: Same here. Possibly add a more generic example?

src/fprime/fbuild/interaction.py
```diff
@@ -69,41 +74,91 @@ def run_impl(deployment: Path, path: Path, platform: str, verbose: bool):
         lines = lines[11:]  # Remove old header
         with open(dest, "a") as file_handle:
             file_handle.write("".join(lines))
-    removals = [path / "mod.mk", path / "Makefile", Path(hpp_src), Path(cpp_src)]
```
Joshua-Anderson, line 72: I think you may still need this step to remove the leftover `.hpp-template` and `.cpp-template` files.

src/fprime/fbuild/interaction.py
```diff
@@ -194,6 +246,7 @@ def new_component(
                 )
             )
             return 0
+        regenerate(cmake_lists_file)
```
Joshua-Anderson: Instead of regenerating for the user, I wonder if you should do what `fprime-util build` does, where it checks if the build directory exists and if it doesn't product an error message with instructions to run generate. Sometimes when the build directory doesn't exist it just means that the user forgot to specify the right platform, so returning an error allows the user to either correct the platform or run generate, whichever is correct for the current situation.

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +270,144 @@ def new_component(
...
+    # Gather inputs to use as context for the port template
+    defaults = {
+        "port_name": "ExamplePort",
+        "short_description" : "Example usage of port",
+        "dir_name" : "example_directory",
+        "arg_number" : 1,
+    }
+    valid_name = False
+    valid_dir  = False
+    invalid_characters = ["#", "%", "&", "{", "}", "/", "\\", "<", ">", "*", "?",
```
Joshua-Anderson: We also check for the same invalid characters when generating components. Do you think it's worth making this a generic function that can be shared so you don't have to duplicate the list of invalid characters and the logic for checking if the input contains invalid characters?

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +270,144 @@ def new_component(
...
+        if values[key] == "":
+            values[key] = defaults[key]
+    return values
+
+def make_namespace(deployment, cwd):
+    # Form the namespace from the path to the deployment
+    namespace_path = cwd.relative_to(deployment)
+    deployment_list = str(deployment).split("/")
+    deployment_dir = deployment_list[-1]
+    whole_path = deployment_dir + "/" + str(namespace_path)
```
Joshua-Anderson: One totally optional tip is it might be worth considering using python's [pathlib](https://docs.python.org/3/library/pathlib.html) for handling path manipulation like this. Whenever I see usage of raw `/` symbols I get a little nervous about they perform under certain edge cases, like with `a//b/c`. Pathlib is nice because it helps handle paths in an OS independent way.

src/fprime/fbuild/interaction.py
```diff
@@ -69,41 +74,91 @@ def run_impl(deployment: Path, path: Path, platform: str, verbose: bool):
...
+    except:
+        os.system("fprime-util generate")
+    os.chdir(currentDir)
+
+
+def add_unit_tests(comp_path):
+    #Creates unit tests and moves them into test/ut directory
+    os.chdir(str(comp_path))
+    if confirm("Would you like to generate unit tests?: "):
+        os.system("fprime-util impl --ut")
```
Joshua-Anderson: Instead of executing the currently running program by calling `os.system`, what about calling the `impl --ut` target from within python by calling the respective target?

src/fprime/fbuild/interaction.py
```diff
@@ -69,41 +74,91 @@ def run_impl(deployment: Path, path: Path, platform: str, verbose: bool):
...
+        os.rename("Tester.hpp", "test/ut/Tester.hpp")
```
Joshua-Anderson: Is there some way for us to adjust `impl --ut` so that it just creates the files in the correct directory instead of moving them?

src/fprime/fbuild/settings.py
```diff
@@ -148,6 +148,9 @@ def load(settings_file: Path, cwd: Path):
             "install_dest": install_dest,
             "environment_file": env_file,
             "environment": environment,
+            "cookiecutter": confparse.get(
+                "fprime", "cookiecutter", fallback="native"
```
Joshua-Anderson: What about using `default` or `builtin` instead of native for the default value?

src/fprime/util/build_helper.py
```diff
@@ -71,7 +71,7 @@ def validate(parsed, unknown):
         }
         cmake_args.update(d_args)
     # Build type only for generate, jobs only for non-generate
-    elif parsed.command not in ["info", "purge", "hash-to-file", "new"]:
+    elif parsed.command not in ["info", "purge", "hash-to-file", "new", "new-port"]:
```
Joshua-Anderson: It looks like the you use `new` for new ports a well. Is `new-port` used?

Joshua-Anderson (changes requested):
This is a great first revision!

I had a bunch of various feedback on the changes, but I'm definitely not an expert on cookiecutter and how we're using using it. Don't be afraid to push back and let me know if you disagree with or are confused by any of my comments!

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/CHANGELOG.rst
aidan-wagner: I think this can be removed, so I'll take it out

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/README.rst
aidan-wagner: I think we have simplified the questions enough that this can probably also be removed

aidan-wagner: I think we have simplified the questions enough that this can probably also be removed. Although, do you think we need some type of readme explaining the cookiecutter to the user?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
aidan-wagner: Ok, that does seem like better coding practice

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/README.rst
Joshua-Anderson: We should definitely add some documentation on `fprime-util new` to the F' documentation! It should definitely be referenced on the [developer workflow page](https://nasa.github.io/fprime/UsersGuide/best/development-practice.html) as well

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
aidan-wagner: Ok, I think that's the only line that uses the path_to_fprime, so I should be able to trim down some other code too!

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/.cookiecutterrc
aidan-wagner: You're right, this doesn't seem like a helpful feature for us

src/fprime/util/build_helper.py
aidan-wagner: That must have been left over from my initial new port template, but I'll remove it now

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
aidan-wagner: Let me know if I'm seeing this wrong, but to me it actually looks like click.termui is only used to define the function warn(), which is never actually called. If this is the case can we just get rid of this import and the function definition?

Joshua-Anderson: Yep, you're right I don't think warn is used at all. Removing it seems like the way to go.

src/fprime/fbuild/interaction.py
aidan-wagner: You're right! I'll add the part of this that we need back in.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
aidan-wagner: These were taken from Sterling's cookicutter template, but I agree that something like "ExampleCommand" would be simpler

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
aidan-wagner: TestChannel makes me think that were testing some piece of code. Is there some reason you chose that instead of ExampleChannel?

Joshua-Anderson: `ExampleChannel` sounds good to me!

src/fprime/fbuild/interaction.py
aidan-wagner: That does seem like a good idea

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/README.rst
aidan-wagner: Where would it be put in the F' documentation? I'm having trouble finding where the other commands are documented

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
aidan-wagner: Ok that makes sense. And I just took the numbers from Sterling's template, so I assumed there was a reason for his numbering, but if not then I'll just set them to 1

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
aidan-wagner: I think that's a good idea, but I also like keeping two examples here because I feel like it shows the optional argument well. Is it okay to just have two generic examples here?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/README.rst
Joshua-Anderson: fprime-util's guide is [here](https://github.com/nasa/fprime/blob/devel/docs/UsersGuide/user/fprime-util.md) and the developer workflow is documented [here](https://github.com/nasa/fprime/blob/devel/docs/UsersGuide/best/development-practice.md). These markdown files are used by github pages to generate the fprime website.

src/fprime/fbuild/settings.py
aidan-wagner: Ok sounds good, but could give a quick explanation to the benefits of this change because I'm not sure I understand the difference?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/README.rst
aidan-wagner: Ok thanks! Should I add this documentation now, or wait until the demo on next Tuesday, in case there are changes?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/.gitignore
aidan-wagner: This does seem redundant, I'll remove it

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/{{cookiecutter.component_name}}.md
aidan-wagner: When you say the autocoded html dictionary, do you mean the sdd.md file, or is there a different one I'm unaware of?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
Joshua-Anderson: It does do a good job of demonstrating the optional argument here, but is this the place to demonstrate that? I feel something like that better belongs in the documentation. I think demonstrating this every time somebody generates a new component is a little excessive, especially if it requires users to remove autogenerated code they're not using.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/{{cookiecutter.component_name}}.md
Joshua-Anderson: The autocoder generates a dictionary for you! Take a look in `build-artifacts/dict`

src/fprime/fbuild/settings.py
Joshua-Anderson: We use native to mean a native build, as in not-cross compiled and you're building for your local computer. Native just doesn't seem to make sense in the context of selecting which cookiecutter template to use

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/README.rst
Joshua-Anderson: I think it's unlikely to change a large amount, so you probably could start working on it if you had the time. I'd prioritize  working on this pull request with the implementation over the documentation right now though

src/fprime/fbuild/settings.py
aidan-wagner: Ok makes sense thanks!

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/{{cookiecutter.component_name}}.md
aidan-wagner: Where is build-artifacts? I can't seem to find it

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
aidan-wagner: That is a good point. Especially because the cookiecutter won't only be used for learning purposes. I'll remove one of them then

src/fprime/fbuild/interaction.py
aidan-wagner: Got this working!

src/fprime/fbuild/interaction.py
aidan-wagner: Do you mean a change to the fprime-util impl --ut command, or just a change to the way cookiecutter uses it?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/{{cookiecutter.component_name}}.md
Joshua-Anderson: It'll be generated whenever you run fprime-util build and it'll be in the same directory that your cmake directories are created in (root of deployment). If you're building the Ref example it'll show up in `Ref/build-artifacts/` directory

src/fprime/fbuild/interaction.py
Joshua-Anderson: Either!

aidan-wagner: I found the target for impl --ut, but I can't seem to find where the code for impl --ut is. Could you point me to where it is?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/{{cookiecutter.component_name}}.md
aidan-wagner: I found the build artifacts directory, but whats inside that is a Linux directory, and then inside that there's folders for commands, channels, events, and then files for the topology. I didn't see any dictionaries for components

src/fprime/fbuild/interaction.py
aidan-wagner: Ok, I changed this function to use pathlib and now it should work independent of the OS!

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/{{cookiecutter.component_name}}.md
Joshua-Anderson: The commands, channels, events are the dictionary! 😄  In those directories you can see html files that contain the autogenerated documentation for events, commands, and telemetry. I think this should replace the manual dictionary that you have to manually keep up to date with the new commands and events you add to your component.

src/fprime/fbuild/interaction.py
Joshua-Anderson: It's part of the autocoder in the F' repo. We could tweak the location the autocoder generates the UT files, but doing that in a backwards incompatible way would be tricky. I think it would be worth updating the autocoder not to generate the unnecessary  `TesterBase` and `GTestBase` files though. @LeStarch any thoughts?

I think you can leave this as is, but try to make sure the this doesn't crash if any of the expected files don't exist (I.E. if we decide not to generate those unnecessary files).

Joshua-Anderson:
@aidan-wagner can you clean up the new lgtm warnings?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/README.rst
```diff
@@ -0,0 +1,177 @@
...
+Cookiecutter: F Prime Component
```
Joshua-Anderson: Did we decide that we were removing this readme?

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
```diff
@@ -21,13 +20,15 @@ def main():
     cwd = Path(os.getcwd())
     deployment = Build.find_nearest_deployment(cwd)
     namespace = make_namespace(deployment, cwd)
-    try:
+    if os.path.isfile("../settings.ini"):
```
Joshua-Anderson: I think that `IniSettings.load` will already check if the provided settings.ini file exists and warn if it doesn't, so you may be able to remove this preliminary check that the files exists.

You'll also probably want to use the path `deployment / "settings.ini"` just in cause the user puts components in  a subdirectory and the settings file isn't at `../settings.ini` relative to the working directory.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
```diff
@@ -66,15 +67,15 @@ def main():
...
-                you will need to add this component to 
-                your build system and then possibly purge and generate your project.
-
-                In addition, the unit test files were not generated.
-                
-                To fix this issue, ensure you have a settings.ini file that 
-                specifies your project_root
-                """
+        No project root was specified in your settings.ini file. This means
+        you will need to add this component to
```
Joshua-Anderson: Very optional nitpick: reflow the string so that the length of this line is consistent.

test/fprime/fbuild/test_settings.py
```diff
@@ -39,6 +39,7 @@ def test_settings():
                 "library_locations": [],
                 "environment_file": full_path("settings-data/settings-empty.ini"),
                 "environment": {},
+                "cookiecutter": "native",
```
Joshua-Anderson: These probably need to changed from native to default to match with code changes.
Also, adding a test case for explicitly setting the cookiecutter template would be awesome.

src/fprime/fbuild/interaction.py
```diff
@@ -69,41 +76,98 @@ def run_impl(deployment: Path, path: Path, platform: str, verbose: bool):
...
+
+
+def regenerate(cmake_list_file):
+    # Ensures build directory exists, then refreshes build cache
+    build_dir = Path(cmake_list_file.parent, "build-fprime-automatic-native")
+    if not os.path.isdir(build_dir):
+        raise CMakeInvalidBuildException(build_dir)
+    handler = CMakeHandler()
+    print("Refreshing cache to include new addition")
+    handler._cmake_refresh_cache(build_dir)
```
Joshua-Anderson: We should rename `_cmake_refresh_cache` to `cmake_refresh_cache`. I think the _ is used to denote the function is private, which is no longer the case.

src/fprime/fbuild/interaction.py
```diff
@@ -69,41 +76,98 @@ def run_impl(deployment: Path, path: Path, platform: str, verbose: bool):
...
+        return True
+    lines.insert(index, addition)
+    with open(list_file, "w") as f:
+        f.write("".join(lines))
+    return True
+
+
+def regenerate(cmake_list_file):
+    # Ensures build directory exists, then refreshes build cache
+    build_dir = Path(cmake_list_file.parent, "build-fprime-automatic-native")
```
Joshua-Anderson: We probably don't want to assume the build directory name. Can you use the same method the rest of fprime-util uses to determine the build directory name? I think it's by instantiating the build class.

src/fprime/fbuild/interaction.py
```diff
@@ -69,41 +76,98 @@ def run_impl(deployment: Path, path: Path, platform: str, verbose: bool):
...
-    if topology_lines:
-        line, text = topology_lines[0]
-        print(
-            "[INFO] Topology inclusion '{}' found on line {}.".format(
-                text.strip(), line + 1
-            )
-        )
+    index = 0
+    line = lines[index]
+    while line != "set(SOURCE_FILES\n":
```
Joshua-Anderson: We might want to make this a regular expression comparison rather than a direct string comparison.

Otherwise we'll fail to match in the case of something like `set( SOURCE_FILES` where a space is added.

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +282,196 @@ def new_component(
...
+        fname = context["port_name"] + "Port" + "Ai.xml"
+        with open(fname, "w") as f:
+            xml_file = TEMPLATE_ENVIRONMENT.get_template("port_template.xml").render(
+                context
+            )
+            f.write(xml_file)
+        if not os.path.isdir(context["dir_name"]):
+            os.mkdir(context["dir_name"])
+
+        os.rename(fname, context["dir_name"] + "/" + fname)
```
Joshua-Anderson: Could you replace this group of path manipulations with pathlib as well?

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +282,196 @@ def new_component(
...
+        PATH = os.path.dirname(os.path.abspath(__file__))
+        TEMPLATE_ENVIRONMENT = Environment(
+            autoescape=False,
+            loader=FileSystemLoader(os.path.join(PATH, "../cookiecutter_templates")),
+            trim_blocks=False,
+        )
+        params = get_port_input()
+        if (
+            settings.get("framework_path") is not None
+            and settings["framework_path"] != "native"
```
Joshua-Anderson: Framework path will always be a path, right? Is there a situation where it could get set to "native"?

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +282,196 @@ def new_component(
...
+    return namespace_formatted
+
+
+def new_port(cwd: Path, deployment: Path, settings: Dict[str, str]):
+    """ Uses cookiecutter for making new ports """
+    try:
+        print("[WARNING] **** fprime-util new is prototype functionality ****")
+        proj_root = None
+        try:
+            proj_root = Path(settings.get("project_root", None))
```
Joshua-Anderson: It might be a bit cleaner to check if project root exists rather than trying to trap an error creating a path with None.

src/fprime/cookiecutter_templates/port_template.xml
```diff
@@ -0,0 +1,14 @@
...
+<?xml-model href="{{path_to_fprime_root}}/../Autocoders/Python/schema/default/component_schema.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
+<interface name="{{port_name}}" namespace="{{namespace}}">
+    <comment>
+    {{short_description}}
+    </comment>
+    <args>
+    {% for i in range(arg_number) %}
+        <arg name="arg {{ i }}" type="U32">
+            <comment>Argument {{ i }}</comment>
+		</arg>
```
Joshua-Anderson: This indentation doesn't look quite right.

src/fprime/cookiecutter_templates/port_template.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<?xml-model href="{{path_to_fprime_root}}/../Autocoders/Python/schema/default/component_schema.rng" type="application/xml" schematypens="http://relaxng.org/ns/structure/1.0"?>
```
Joshua-Anderson: We're no longer using xml models, so you can remove this line.

Joshua-Anderson (changes requested):
Getting close! The revisions looked good, just I found a few more things looking through the code.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/README.rst
aidan-wagner: Right! I must've just forgotten to actually remove it

test/fprime/fbuild/test_settings.py
aidan-wagner: Ok! I made this test case, but I get the same error with or without my test case when running pytest: that the code can't find the fprime root.

Joshua-Anderson: Yeah, you have to set the python root before running tests. Try `export PYTHONPATH=$PWD/src` when you're in the repository.

src/fprime/fbuild/interaction.py
aidan-wagner: I think I may actually be able to get rid of this block because framework path is only used in the line that gets the schema in the port_template.xml

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
```diff
@@ -0,0 +1,190 @@
...
+        lines = f.readlines()
+        f.close()
+    with open(filename, "w") as f:
+        for line in lines:
+            if line != removal:
+                f.write(line)
+        f.close()
+
+
+def update_sdd():
```
Joshua-Anderson: I'm wondering if these detailed pre-command/event/channel descriptions in the SDD are necessary given that we have a dictionary. I like keeping commannd descriptions in one place because if they're in multiple places they inevitably become out of sync.

Since we don't have ports in the dictionary, ports descriptions should probably remain in the SDD.

What are your thoughts @LeStarch?

src/fprime/fbuild/interaction.py
```diff
@@ -69,41 +74,100 @@ def run_impl(deployment: Path, path: Path, platform: str, verbose: bool):
...
-        print(
-            "[INFO] Topology inclusion '{}' found on line {}.".format(
-                text.strip(), line + 1
-            )
-        )
+    index = 0
+    while re.search("set\(\s*SOURCE_FILES", lines[index]) is None:
+        index += 1
+    index += 1
+    while re.search("CMAKE_CURRENT_LIST_DIR", lines[index]):
```
Joshua-Anderson: If you're not using regex patterns, it might be simpler to go with a substring search instead:

```
if "CMAKE_CURRENT_LIST_DIR" in lines[index]:
```

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +279,215 @@ def new_component(
...
+        "dir_name": ".",
+        "namespace": namespace,
+        "arg_list": [],
+    }
+    valid_name = False
+    valid_dir = False
+    valid_namespace = False
+    args_done = False
+    arg_list = []
+    while not valid_name:
```
Joshua-Anderson: optional recommendation: You might be able to simplify the code and remove some duplication by refactoring the user input prompt into a function.

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +279,215 @@ def new_component(
...
+
+        PATH = os.path.dirname(os.path.abspath(__file__))
+        TEMPLATE_ENVIRONMENT = Environment(
+            autoescape=False,
+            loader=FileSystemLoader(os.path.join(PATH, "../cookiecutter_templates")),
+            trim_blocks=False,
+        )
+        params = get_port_input(deployment.name)
+
+        context = {
```
Joshua-Anderson: Isn't this context dict identical to the params dictionary?

Joshua-Anderson (changes requested):
This is getting really close! Just a few small recommendations.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
LeStarch: @Joshua-Anderson We typically place these in the SDD.  It is "best" to generate an aux documentation file from the dictionary that can be linked to....but that doesn't work anymore.

src/fprime/fbuild/interaction.py
aidan-wagner: Whoops, you're right!

src/fprime/fbuild/interaction.py
aidan-wagner: I agree, this could be much simpler

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
aidan-wagner: @Joshua-Anderson and @LeStarch So should I just leave it as is then, or is there something I should change?

Joshua-Anderson: > It is "best" to generate an aux documentation file

Are the html dictionary files not the aux documentation files?
Right now this SDD section is identical to the HTML dictionary docs so it feels a bit redundant.

aidan-wagner: So do you think it would be best to remove this from the HTML documentation files?

Joshua-Anderson: No, the html files are autogenerated from the F' dictionary, so we get them for "free". We're definitely not removing them, I'm just clarifying with Michael that they're not sufficient by themselves, since it seems like the information overlaps with the SDD command and event descriptions

Joshua-Anderson: Let's keep this section in the SDD. You can't build per-component dictionaries so having a high level dictionary in the SDD is the most convenient place to reference a component's ports/commands/channels/events/etc.

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/hooks/post_gen_project.py
```diff
@@ -0,0 +1,190 @@
...
+    ports = "## Port Descriptions\n| Name | Description |\n"
+    if (
+        "{{cookiecutter.component_kind}}"  # lgtm [py/comparison-of-constants]  lgtm [py/constant-conditional-expression]
```
Joshua-Anderson: I just thought of an idea that might allow you to remove all these manual lgtm warning suppression comments. What you make all cookiecutter inputs to this function parameters, then call the function with the cookiecutter options.

Ex:

```
def update_sdd(component_kind):
    ...

update_sdd("{{cookiecutter.component_kind}}")
```

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +279,194 @@ def new_component(
...
+    dir_name = get_valid_input("Directory Name [{}]: ".format(defaults["dir_name"]))
+    namespace = get_valid_input("Port Namespace [{}]: ".format(defaults["namespace"]))
+    while not args_done:
+        if arg_list == []:
+            add_arg = confirm("Would you like to add an argument?: ")
+        else:
+            add_arg = confirm("Would you like to add another argument?: ")
+        if add_arg:
+            arg_name = input("Argument name: ")
+            arg_type = input("Argument type: ")
```
Joshua-Anderson: It might be nice to prompt a user to pick from a list of default types: U8/I8/F32/F64/BOOL/etc. If a user wants to use a custom type there could be an other type that prompts for a raw string.

I think this could be useful for users unfamiliar with F' to know what the valid options for argument type are. What do you think?

Joshua-Anderson (commented):
Just a couple minor comments!

src/fprime/fbuild/interaction.py
aidan-wagner: Yeah, I like that idea. How many types do we want to list before you think it would be too many options?

Joshua-Anderson: I'd list them all, since only listing some of them would cause confusion as well. I think it's about a dozen options, so hopefully we can fit it in a couple lines.

LeStarch: What about custom types?

aidan-wagner: What if we just include the default types in the prompt, but then allow the user to enter whatever type they want? That way new users would still be able to see what some valid options are.

Joshua-Anderson: Another option would be to have a list of types and then an `other` option. If the user selects other, prompt the user to enter their type. This protects against typos but may be kinda overkill. Your recommendation is a good approach as well, go with whichever you prefer.

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +279,200 @@ def new_component(
...
+        if arg_list == []:
+            add_arg = confirm("Would you like to add an argument?: ")
+        else:
+            add_arg = confirm("Would you like to add another argument?: ")
+        if add_arg:
+            arg_name = get_valid_input("Argument name: ")
+            arg_type = get_valid_input(
+                "Argument type (Valid primitive types are: BOOL, I8, I16, "
+                + "I32, U8, U16, U32, F32, F64, NATIVE_INT_TYPE, NATIVE_UNIT_TYPE, and POINTER_CAST. "
+                + "You may also use ENUM or your own user-defined types): "
```
Joshua-Anderson: I think FPP is trying to get rid of inline enums, so we not want to list ENUM as an option. @LeStarch, what are your thoughts?

LeStarch: We should not include ENUM as these are user-defined types.

aidan-wagner: Ok, I'll remove it

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/{{cookiecutter.component_name}}ComponentAi.xml
```diff
@@ -0,0 +1,130 @@
...
+
+    <comment>{{cookiecutter.component_short_description}}</comment>
+
+    <ports>
+        <!-- Input ports must be of kind "sync_input" or "async_input" or "guarded_input" -->
+        <!-- Input ping port -->
+{% if cookiecutter.commands == "yes" %}
+        <!-- Command port definitions: command input receives commands, command reg out, and response out are
+        ports used to register with the command dispatcher, and return responses to it -->
+        <port name="cmdIn" data_type="Fw::Cmd"  kind="input" role="Cmd"    max_number="1">
```
Joshua-Anderson: Super minor nitpick, but there's extra spaces after `role="Cmd"`

LeStarch:
@aidan-wagner, you should run the formatter:

```
pip install black
cd fprime-tools
black .
git add/commit
```

Joshua-Anderson:
It looks like the F' and Cookiecutter are fighting.

If I remember correctly, we decided that component filenames should be `MyExample.cpp`. The problem is that the autocoder generates includes that expect the old naming style (`MyExampleComponentImpl.hpp`). Because we want to support a wide variety of versions of F', we can't just fix this in the F' framework autocoder. The best solution I can think of is slightly ugly... detect if there's any incorrect filenames in the include path and re-write them. I.E. replace `#include <Ref/MyExample/MyExampleComponentImpl.hpp>` with  `#include <Ref/MyExample/MyExample.hpp>`

The alternative would be to revert to the old naming convention of  `MyExampleComponentImpl`

```
[ 96%] Building CXX object Ref/MyExample/CMakeFiles/Ref_MyExample.dir/MyExample.cpp.o
/Users/joshuaa/p/fprime-oss/Ref/MyExample/MyExample.cpp:7:10: fatal error: 'Ref/MyExample/MyExampleComponentImpl.hpp' file not found
#include <Ref/MyExample/MyExampleComponentImpl.hpp>
```

aidan-wagner:
@LeStarch I ran `black .` before I pushed the code, and when running it again, it doesn't change the code. Is there another way to fix this line? Or are the format rules somewhere so I could fix it by hand?

aidan-wagner:
@Joshua-Anderson Yeah, this is something that @LeStarch and I realized. I thought I remembered him saying that we would make this change in fprime itself, but if not then I can make a change to the cookiecutter

Joshua-Anderson:
@aidan-wagner Are you running a newer version of black than CI? We may need to update our black formatter version in CI

LeStarch:
...yes we should fix this in F´.  Is this fix a blocker for this code?

Joshua-Anderson:
@LeStarch Are we okay with fprime-util new not working with older versions of F' (like 2.0)? It seems like it's easy enough (if a it bit ugly) to support backwards compatibility for now.

LeStarch:
@Joshua-Anderson @aidan-wagner I'd phrase it this way: it isn't worth losing backward compatibility over this.   I think this should be a fairly straight-forward fix, yes?

1. open file and read bytes
2. bytes.replace("ComponentImpl.hpp" ".hpp")
3. write bytes out.

Thoughts?

Joshua-Anderson:
Yeah, that would be my recommendation too. A find and replace is a bit ugly but I think backwards compatibility is worth it. We should definitely fix it in the autocoder though.

aidan-wagner:
@Joshua-Anderson @LeStarch Ok, that makes sense. I can add that in. For the formatting, I tried installing with the same command( so the same version) that the format check on github uses and still no luck.

Joshua-Anderson:
It looks like there's still a comment that redirect to sterling's cookiecutter template. I think we can remove the link to documentation for now.

Joshua-Anderson:
`[INFO] Cookiecutter source: /Users/joshuaa/p/fprime-tools/src/fprime/fbuild/../cookiecutter_templates/cookiecutter-fprime-component`

This doesn't seem super user friendly. What if we say `Cookiecutter source: using builtin` if the user doesn't override the cookiecutter source?

Joshua-Anderson:
`Add component MyExample to /Users/joshuaa/p/fprime-oss/Ref/CMakeLists.txt at end of file??`Are the double `?` marks a typo?

src/fprime/fbuild/interaction.py
```diff
@@ -216,3 +296,200 @@ def new_component(
...
+    namespace = get_valid_input("Port Namespace [{}]: ".format(defaults["namespace"]))
+    while not args_done:
+        if arg_list == []:
+            add_arg = confirm("Would you like to add an argument?: ")
+        else:
+            add_arg = confirm("Would you like to add another argument?: ")
+        if add_arg:
+            arg_name = get_valid_input("Argument name: ")
+            arg_type = get_valid_input(
+                "Argument type (Valid primitive types are: BOOL, I8, I16, "
```
Joshua-Anderson: BOOL actually doesn't look like it's a valid F' type. Not sure how it wound up in the docs. It can probably removed from this list, at least for now

Joshua-Anderson:
`Add component MyExample to /Users/joshuaa/p/fprime-oss/Ref/CMakeLists.txt at end of file?` we don't actually add the component to the end of the file. Maybe we should just ask `Add component MyExample to /Users/joshuaa/p/fprime-oss/Ref/CMakeLists.txt`?

Joshua-Anderson (approved):
🎉

LeStarch:
🎉

PR #6, opened 2021-06-11
Add Black Formatter

Piphi5:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |
|**_Documentation Included (y/n)_**| y  |

---
## Change Description

Adds the [black formatter](https://github.com/psf/black) as an additional action that checks the code style. 

## Rationale

It provides an easy-to-use and reliable tool to ensure that the code is formatted to Python PEP 8 guidelines.

The package can be installed with:
`pip install black`
Then it can be run on the entire directory with `black ./`

## Future Work
NA

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -1,0 +1,17 @@
+name: Format
+
+on: [push, pull_request]
+jobs:
+  format:
+      name: Format
+      runs-on: ubuntu-latest
+      steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Check formatting
+        run: |
+          pip install black
+          black --check --diff ./
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -13,5 +13,5 @@ jobs:
           python-version: 3.7
       - name: Check formatting
         run: |
-          pip install black
+          pip install black==21.6b0
           black --check --diff ./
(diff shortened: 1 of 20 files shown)


.github/workflows/format.yml
```diff
@@ -0,0 +1,17 @@
...
+      runs-on: ubuntu-latest
+      steps:
+      - uses: actions/checkout@v2
+      - name: Setup Python 3.7
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.7
+      - name: Check formatting
+        run: |
+          pip install black
```
Joshua-Anderson: I might pin black to a specific version - that can avoid an annoying edge case where if a new black version is released and reformats our code and breaks CI without us changing anything. I don't think we pin it in our setup.py file either which we may want to correct.

Joshua-Anderson (commented):
Thanks for the contribution! I'm pro setting up CI to enforce code formatting, but it does make things a bit more difficult for new contributors. What do you think @LeStarch?

Piphi5:
>  it does make things a bit more difficult for new contributors.

The nice thing about black is that it has a cli script that will automatically format the files, so at worst, new contributors would have to just run the cli script and push the updated files again.

.github/workflows/format.yml
Piphi5: I'm not sure we should pin it in the setup.py file since we're not calling on black within the fprime-tools module itself, but I've added it to the CI.

LeStarch:
I think it will be just as inconvenient as `spelling` and spelling has been a major boon.  Thus I think we should include it.

LeStarch (approved):

Piphi5:
@LeStarch Should I add this to the other repos as well?

Joshua-Anderson:
@Piphi5 Go for it!

PR #18, opened 2021-08-17
Integration tests

aidan-wagner:
| | |
|:---|:---|
|**_Originating Project/Creator_**|Aidan Wagner |
|**_Affected Component_**| Tests |
|**_Related Issue(s)_**|  |

---
## Change Description

Added integration tests to maintain fprime-tools backwards compatibility

## Rationale

To protect fprime-tools features from being broken by new code

## Future Work

Could test other features besides generate and build

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -3,7 +3,7 @@ name: Integration Tests
 on: [pull_request, push]
 
 jobs:
-  integration_test:
+  fprime-1.5-integration:
 
     runs-on: ubuntu-latest
     strategy:
@@ -24,8 +24,12 @@ jobs:
       - name: clone fprime
         run: |
           git clone -b ${{ matrix.fprime-version }} https://github.com/nasa/fprime.git
+          git clone https://github.com/fprime-community/gps-tutorial.git
       - name: Test generation and building
         run: |
           cd fprime/Ref
           fprime-util generate
           fprime-util build
+          cd ../..
+          mv fprime gps-tutorial
+          cd gps tutorial
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -24,7 +24,7 @@ jobs:
       - name: clone fprime and gps-tutorial
         run: |
           git clone -b ${{ matrix.fprime-version[0] }} https://github.com/nasa/fprime.git
-          git clone -r https://github.com/fprime-community/gps-tutorial.git
+          git clone --recurse-submodules https://github.com/fprime-community/gps-tutorial.git
       - name: Checkout correct gps-tutorial
         working-directory: gps-tutorial
         run: git checkout ${{ matrix.fprime-version[1] }}
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -26,7 +26,9 @@ jobs:
           git clone -b ${{ matrix.fprime-version }} https://github.com/nasa/fprime.git
           git clone https://github.com/fprime-community/gps-tutorial.git
       - name: check gps-tutorial version
-        if: ${{ matrix.fprime-version == 'NASA-v1.5.0' || matrix.fprime-version == 'NASA-v1.5.1' || matrix.fprime-version == 'NASA-v1.5.2' || matrix.fprime-version == 'NASA-v1.5.3'}}
+        if: |
+          ${{ matrix.fprime-version == 'NASA-v1.5.0' || matrix.fprime-version == 'NASA-v1.5.1' || 
+          matrix.fprime-version == 'NASA-v1.5.2' || matrix.fprime-version == 'NASA-v1.5.3'}}
         run: |
           cd gps-tutorial
           git checkout 99cef07f4b43aebd4d3e9298081655b33be859b4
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,49 @@
...
+on: [pull_request, push]
+
+jobs:
+  fprime-integration:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
+        fprime-version: [NASA-v1.5.0, NASA-v1.5.1, NASA-v1.5.2, NASA-v1.5.3, v2.0.0, devel]
```
Joshua-Anderson: I'm wondering if we need to integration test every patch release. Instead of testing 1.5.0-1.5.3 for example, I think we could probably just test 1.5.4. Whenever a new patch release comes up we can update these integration tests to test against the latest patch. What are your thoughts @LeStarch?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,49 @@
...
+          matrix.fprime-version == 'NASA-v1.5.1' || 
+          matrix.fprime-version == 'NASA-v1.5.2' || 
+          matrix.fprime-version == 'NASA-v1.5.3'
+        run: |
+          cd gps-tutorial
+          git checkout 99cef07f4b43aebd4d3e9298081655b33be859b4
+          cd ..
+      - name: Test generation and building on Ref
+        run: |
+          cd fprime/Ref
```
Joshua-Anderson: Instead of all this manual directory changing, github actions allows you to set a working directory. This might clean things up a little bit.

Ex:

```yaml
- name: Test generation and building on Ref
   working-directory: fprime/Ref
   run: |
          fprime-util generate
          fprime-util build
```

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,49 @@
...
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -e .
+      - name: clone fprime and gps-tutorial
+        run: |
+          git clone -b ${{ matrix.fprime-version }} https://github.com/nasa/fprime.git
+          git clone https://github.com/fprime-community/gps-tutorial.git
+      - name: Revert gps-tutorial version if needed
+        if: |
```
Joshua-Anderson: I wonder if there's any better way to do this. Maybe some sort of map that maps F' version to gpsapp version?

.github/workflows/integration-tests.yml
LeStarch: I say we run against what is actively in use:

NASA-v1.5.3, v2.0.0, devel

aidan-wagner: Sounds good!

.github/workflows/integration-tests.yml
aidan-wagner: Ok, I think I found a better way

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+on: [pull_request, push]
+
+jobs:
+  fprime-integration:
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.6, 3.7, 3.8, 3.9]
+        fprime-version: [[NASA-v1.5.3, 99cef07], [v2.0.0, 521516c] , [devel, 521516c]]
```
Joshua-Anderson, line 12: That's really cool! @LeStarch would you be okay with tagging the gpsapp so we don't have to specify random commit hashes here? Also could you add a commit specifying that this tuple is [fprime version, gpsapp version] so that there's no confusion.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,42 @@
...
+          git clone https://github.com/fprime-community/gps-tutorial.git
+      - name: Checkout correct gps-tutorial
+        working-directory: gps-tutorial
+        run: git checkout ${{ matrix.fprime-version[1] }}
+      - name: Test generation and building on Ref
+        working-directory: fprime/Ref
+        run: |
+          fprime-util generate
+          fprime-util build
+      - name: Move fprime into gps-tutorial
```
Joshua-Anderson: The GPS app should already be embedded a tagged version of fprime. I say we just checkout the GPS app and just use whatever fprime commit it includes in it's submodule.

Joshua-Anderson:
@LeStarch @aidan-wagner Do you think we should test that `fprime-util check` works or do you think that that's overkill and that if the build works unit tests should work too?

.github/workflows/integration-tests.yml
aidan-wagner: Got it!

.github/workflows/integration-tests.yml
aidan-wagner: Now that we're just using the fprime version that comes with the gps-tutorial version, how do we want to layout the tests? The fprime versions and the gps-tutorial versions aren't necessarily linked anymore so do we still want to test the older gps-tutorial?

LeStarch: @Joshua-Anderson tag away!

LeStarch:
@Joshua-Anderson `fprime-util check` should be evaluated for:

1. Running inside a directory
2. Running with --all in the root.

Perhaps we need a very simple deployment as these are starting to get pretty dense testing wise.

LeStarch (approved):

PR #27, opened 2021-12-06
lestarch: adding in BUILD_TESTING setup

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

For new versions of fprime (v3) this uses properly the `CMAKE_BUILD_TYPE` argument.  Users can now set this argument in `fprime-util` as expected, but it is set to DEBUG when a testing run is selected with `--ut`.

This will also set `BUILD_TESTING=ON` with the testing run `--ut` so as to generate the unit tests.

## Changes to check:
- Post v3, `BUILD_TESTING=On` and `CMAKE_BUILD_TYPE=Debug` should be used when build type is BUILD_TESTING.
- Pre v3, `CMAKE_BUILD_TYPE=Testing` is used
- `.load` functions have removed the initial path argument
- rebuild cache function now tries quick, and upon exception, runs full

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -443,8 +443,20 @@ class Build:
             cmake_args["FPRIME_LIBRARY_LOCATIONS"] = ";".join(
                 cmake_args["FPRIME_LIBRARY_LOCATIONS"]
             )
-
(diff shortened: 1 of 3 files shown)


kevin-f-ortega (approved):
Looks good to me.

PR #42, opened 2022-01-13
lestarch: added overridable build cache to fprime-util

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Adds in the ability to add a specific build cache using the `--build-cache` flag.  This way any number of configurations can be maintained by hand.

e.g.
```
fprime-util generate --build-cache my_fancy_build -DSOME_SETTING=ON -DSOME_OTHER_SETTING=OFF
```

@vietjtnguyen does this meet your recommendation?  Rather than managing suffixs I figured I'd put in the full explicit override so the user can specify any build cache, even one made by hand.

### DIFF
--- a/src/fprime/common/models/serialize/numerical_types.py
+++ b/src/fprime/common/models/serialize/numerical_types.py
@@ -44,6 +44,7 @@ class NumericalType(ValueType, abc.ABC):
     @abc.abstractmethod
     def get_serialize_format():
         """Gets the format serialization string such that the class can be serialized via struct"""
+        raise NotImplementedError("get_serialize_format")
 
     def serialize(self):
         """Serializes this type using struct and the val property"""
--- a/src/fprime/fbuild/cli.py
+++ b/src/fprime/fbuild/cli.py
@@ -8,7 +8,7 @@ target operations.
 import argparse
 from pathlib import Path
 from typing import Dict, List, Tuple, Callable
-from fprime.fbuild.builder import Target, BuildType, Build, InvalidBuildCacheException
+from fprime.fbuild.builder import Target, BuildType, Build
 from fprime.fbuild.interaction import confirm
 
 
--- a/src/fprime/fbuild/cli.py
+++ b/src/fprime/fbuild/cli.py
@@ -45,7 +45,7 @@ def run_fbuild_cli(
         make_args: arguments to the make system
     """
     if parsed.command == "generate":
-        build.invent(parsed.platform)
+        build.invent(parsed.platform, build_dir=parsed.build_cache)
         toolchain = build.find_toolchain()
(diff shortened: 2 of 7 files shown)


LeStarch:
@Joshua-Anderson would it be possible to have you review this PR?

vietjtnguyen:
@LeStarch This would be perfect.

src/fprime/common/models/serialize/numerical_types.py
```diff
@@ -44,6 +44,7 @@ def getSize(cls):
     @abc.abstractmethod
     def get_serialize_format():
         """Gets the format serialization string such that the class can be serialized via struct"""
+        raise NotImplementedError("get_serialize_format")
```
Joshua-Anderson, line 47: Is this a part of the build cache changes?

LeStarch: No, it fixes an outstanding static analysis issue.  This method shouldn't be called (overridden in all implementations) so it shouldn't harm to properly raise this exception.

src/fprime/fbuild/cli.py
```diff
@@ -45,36 +45,33 @@ def run_fbuild_cli(
...
-                        purge_build.purge_install()
+        for purge_build in Build.get_build_list(build, parsed.build_cache):
             print(
                 f"[INFO] {parsed.command.title()} build directory at: {purge_build.build_dir}"
             )
             if parsed.force or confirm("Purge this directory (yes/no)?"):
                 purge_build.purge()
+            install_dir = purge_build.install_dest_exists()
+            if (
+                purge_build.build_type != BuildType.BUILD_CUSTOM
```
Joshua-Anderson, line 64: Wouldn't you still want to prompt about remove the build artifacts directory on any run of purge? Custom build types are still going to install to the same installation directory, aren't they?

Joshua-Anderson (dismissed):
Looks great!

I had one comment which probably comes down to a philosophical difference, so feel free to merge if you disagree.

I would encourage you to add a really quick test case for this to the fprime-util integration tests since if this isn't actively tested it seems easy to accidentally break, but not required.

src/fprime/fbuild/cli.py
LeStarch: Custom builds can override the build artifacts directory.  Thus, using them should require users to purge-by-hand.

LeStarch:
@Joshua-Anderson I made a few more changes.  Thoughts?  This fixes `info` `hash-to-file` so that it works with custom cache, and adds an integration test that tests for the custom build setup!

Joshua-Anderson (approved):
Looks great, thanks for adding the integration test!

PR #53, opened 2022-04-09
Improve dictionary updating

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---

## Change Description

This PR includes the following changes:
- Added a single value to the dictionary directly rather than using update() [ref](https://docs.sourcery.ai/refactorings/simplify-dictionary-update/).
- Changed dictionary assignments and updates to use the union operator [ref](https://docs.sourcery.ai/refactorings/dict-assign-update-to-union/).

## Rationale

Adding an entry directly to a dictionary is easier than using the update() method. It also eliminates the overhead of creating a new dictionary and calling a method, which should improve performance a bit.

Look out for the union operator for dictionaries, it was added in **Python 3.9** ([PEP 584](https://peps.python.org/pep-0584/)) and provides a concise way to merge dictionaries without using the copy/update statements.

(void)

## Future Work

(void)

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -457,12 +457,10 @@ class Build:
             v3_autocoder_directory.exists()
             and self.build_type == BuildType.BUILD_TESTING
         ):
-            cmake_args.update({"BUILD_TESTING": "ON"})
-            cmake_args.update(
-                {"CMAKE_BUILD_TYPE": cmake_args.get("CMAKE_BUILD_TYPE", "Debug")}
-            )
+            cmake_args["BUILD_TESTING"] = "ON"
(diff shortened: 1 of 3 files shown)


LeStarch:
Our code needs to support active-life python versions.  Thus the `|=` operator in this PR will need to be reverted as it is only supported as near as I can tell in Python 3.9+.

LeStarch:
I look here to see the versions that are still active: https://endoflife.date/python

ThibFrgsGmz:
@LeStarch Um, I don't know how to do that, especially since I deleted my original fork... I think it would be better to close this PR, no?

LeStarch:
Yeah....looks like it.

ThibFrgsGmz:
@LeStarch New PR without the ``|=`` is https://github.com/fprime-community/fprime-tools/pull/69

PR #70, opened 2022-05-22
Replace empty collection comparisons w/ a more idiomatic unary operator

ThibFrgsGmz:
| | |
|:---|:---|
|**_Originating Project/Creator_**| @ThibFrgsGmz |
|**_Affected Component_**| fbuild |
|**_Affected Architectures(s)_**| (void )|
|**_Related Issue(s)_**| (void) |
|**_Builds Without Errors (y/n)_**| y |
|**_Unit Tests Pass (y/n)_**| see CI results |

---
## Change Description

This PR aims to replace the comparison of empty collections with a more idiomatic unary operator.

## Rationale

This refactoring follows the [PEP 8 programming recommendations](https://peps.python.org/pep-0008/#programming-recommendations), in particular the following recommendation: ``For sequences, (strings, lists, tuples), use the fact that empty sequences are false`` in the context of comparisons.

(void)

## Future Work

(void)

### DIFF
--- a/src/fprime/fbuild/cmake.py
+++ b/src/fprime/fbuild/cmake.py
@@ -107,7 +107,7 @@ class CMakeHandler:
             module = self.get_cmake_module(path, build_dir)
             cmake_target = (
                 module
-                if target == ""
+                if not target:
                 else (f"{module}_{target}".lstrip("_") if not top_target else target)
             )
         run_args = ["--build", build_dir]
@@ -556,7 +556,7 @@ class CMakeHandler:
                 appendable.append(line)
                 # Streams are EOF when the line returned is empty. Once this occurs, we are responsible for closing the
                 # stream and thus closing the select loop. Empty strings need not be printed.
-                if line == "":
+                if not line:
                     key.fileobj.close()
                     continue
                 # Forwards output to screen.  Assuming a PTY is used, then coloring highlights should be automatically
--- a/src/fprime/fbuild/cmake.py
+++ b/src/fprime/fbuild/cmake.py
@@ -107,7 +107,7 @@ class CMakeHandler:
             module = self.get_cmake_module(path, build_dir)
             cmake_target = (
                 module
-                if not target
+                if target == ""
                 else (f"{module}_{target}".lstrip("_") if not top_target else target)
             )
         run_args = ["--build", build_dir]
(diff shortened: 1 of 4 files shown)


ThibFrgsGmz:
Hum LGTM does not seem to like PEP8 if I understand correctly?

src/fprime/fbuild/settings.py
```diff
@@ -54,7 +54,7 @@ def read_safe_path(
         all_paths = parser.get(section, key, fallback="").split(":")
         expanded = []
         for path in all_paths:
-            if path == "" or path is None:
```
LeStarch, line 57: `if not path` is sufficient here....as a None is also falsey.

src/fprime/fbuild/interaction.py
```diff
@@ -391,7 +391,7 @@ def get_port_input(namespace):
 
     # Fill in blank values with defaults
     for key in values:
-        if values[key] == "":
```
LeStarch: I am hesitant here.  If somehow a value is correctly set to the literal `False` the new code will override it, and the old code would keep it.  This only works if `values[keys]` is always a string.  I am not confident that assumption will always hold.

src/fprime/fbuild/cmake.py
```diff
@@ -107,7 +107,7 @@ def execute_known_target(
             module = self.get_cmake_module(path, build_dir)
             cmake_target = (
                 module
-                if target == ""
```
LeStarch: This one is another where I am hesitant.  The purpose of this call was to check for exactly empty string.  In this system targets are of the form:

1. target, for top level targets
2. module_target, for module targets
3. module, special case, only when target is exactly ""

src/fprime/fbuild/settings.py
ThibFrgsGmz: Correct!

src/fprime/fbuild/cmake.py
ThibFrgsGmz: Corrected

src/fprime/fbuild/interaction.py
ThibFrgsGmz: Corrected!

LeStarch (approved):

PR #74, opened 2022-05-25
Turns a method into a staticmethod because no `self` is used.

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**| (void) |

---
## Change Description

This PR aims to transform some methods into static methods when no self is used.

## Rationale

These methods do not use their bound instance.
It is advisable to decorate this type of method with the ``@staticmethod`` decorator, so that Python does not have to instantiate a linked method for each instance of this class, thus saving memory and computation. 

(void)

## Future Work

(void)

### DIFF
--- a/src/fprime/fbuild/target.py
+++ b/src/fprime/fbuild/target.py
@@ -39,7 +39,8 @@ class ExecutableAction(ABC):
         """Set scope of this action"""
         self.scope = scope
 
-    def is_supported(self, builder: "Build", context: Path):
+    @staticmethod
+    def is_supported(builder: "Build", context: Path):
         """Is supported by the list of build target names
 
         Checks if the build target names supplied will support this target. Is overridden by subclasses.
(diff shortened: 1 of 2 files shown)


src/fprime/fbuild/target.py
```diff
@@ -39,7 +39,8 @@ def __init__(self, scope: TargetScope):
         """Set scope of this action"""
         self.scope = scope
 
-    def is_supported(self, builder: "Build", context: Path):
```
LeStarch: My recollection when writing this file was that the subclasses overriding this need to access to self. Hence, the base method doesn't need a self-reference, but the overriding methods do.  I'd revert all three methods here.

LeStarch (changes requested):
See the comment on target.py

src/fprime/fbuild/target.py
ThibFrgsGmz: OK, actually, I suspected something like that and was hesitant on this base class. I will correct tomorrow, unless you want to have the pleasure of doing it ;)

ThibFrgsGmz: Done.

LeStarch (approved):

LeStarch (approved):

PR #79, opened 2022-06-05
Refactoring by swapping if-expr to avoid negation

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

This PR aims at swapping some "if" and "else" branches of two if expression to remove a negation.

Misc:
- remove two unneeded f-string format.
- improve dict creation to be more concide and improve perf.

## Rationale

Negative conditions are harder to read than positive conditions, so it is best to avoid them if possible. By swapping the if and else conditions, we can reverse the condition and make it positive.

void

## Future Work

void

### DIFF
--- a/src/fprime/fbuild/cmake.py
+++ b/src/fprime/fbuild/cmake.py
@@ -108,7 +108,9 @@ class CMakeHandler:
             cmake_target = (
                 module
                 if target == ""
-                else (f"{module}_{target}".lstrip("_") if not top_target else target)
+                else target
+                if top_target
+                else f"{module}_{target}".lstrip("_")
             )
         run_args = ["--build", build_dir]
         environment = {} if environment is None else copy.copy(environment)
--- a/src/fprime/fbuild/cmake.py
+++ b/src/fprime/fbuild/cmake.py
@@ -111,7 +111,7 @@ class CMakeHandler:
                 cmake_target = target
             else:
                 cmake_target = f"{module}_{target}".lstrip("_")
-            
+
         run_args = ["--build", build_dir]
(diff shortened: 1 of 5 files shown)


src/fprime/fbuild/cmake.py
```diff
@@ -108,7 +108,9 @@ def execute_known_target(
             cmake_target = (
                 module
                 if target == ""
-                else (f"{module}_{target}".lstrip("_") if not top_target else target)
+                else target
```
LeStarch: I'd prefer to keep the redundant parenthesis here. They may not be necessary, but it is really hard for me to read because there is not a great association between the items in the last clause.  It reads to me like:

```
module if target == "" else target ....noise...
```

Perhaps this means this statement should be a full if statement:

```
if target == ""
    cmake_target = module
elif top_target:
    cmake_target = target
else:
    cmake_target = f"{module}_{target}".lstrip("_")
```

It is more verbose, but a whole lot easier to understand than a nested ternary.  Thoughts?

ThibFrgsGmz: I admit that it can be more complicated to refactor it later. I'll change according to your suggestion.

ThibFrgsGmz:
@LeStarch Note: I've almost finished my first phase of refactoring fprime-tools.

I'm going to start to seriously work on the fprime one. I noticed that PRs take much longer to be accepted on the main repo.

Is there any particular reason? I have a lot of modification on the python autocoder so it's taking a lot longer.

LeStarch:
Ahhh, I found the question!  I answered in one of your PRs.

LeStarch (approved):

PR #88, opened 2022-06-15
Use set comprehension directly instead of calling set()

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

The goal of this PR is to use ``set`` comprehensions directly instead of calling ``set()``.

## Rationale

The Pythonic way to create a set from a generator is to use comprehensions. It is not necessary to use set around a generator expression to get an object of that type, since there are understandings for those types.
Using comprehensions instead of methods is slightly shorter.

void

## Future Work

void

### DIFF
--- a/src/fprime/util/versioning.py
+++ b/src/fprime/util/versioning.py
@@ -32,7 +32,7 @@ def get_version(package: str, requirements: Path):
         )
 
     # Collapse versions that match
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
Looks good, confirmed this feature is in 3.7+

PR #89, opened 2022-06-15
Remove explicit inheritance of `object` class in `HelpText` class

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

The purpose of this PR is to remove the explicit inheritance of the ``HelpText`` class from the ``object`` class. 

## Rationale

In Python 3, the inheritance of the ``object`` class is implicit, and can therefore be safely removed from the bases. [ref](https://stackoverflow.com/questions/4015417/why-do-python-classes-inherit-object)

void

## Future Work

void

### DIFF
--- a/src/fprime/util/help_text.py
+++ b/src/fprime/util/help_text.py
@@ -276,7 +276,7 @@ and components in fprime. The code has not been updated to use FPP models. Pleas
 }
 
 
-class HelpText(object):
+class HelpText:
     """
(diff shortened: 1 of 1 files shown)


LeStarch:
Code change looks good.  Not sure why this one matters.  Is it more Pythonic to omit this?

ThibFrgsGmz:
That's it, since Python 3 it would be useless to mention it.

LeStarch (approved):

PR #93, opened 2022-07-02
Simplify `if` expression by using `or`

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |

---
## Change Description

The purpose of this PR is to replace one `if` expression that compares to a target with `or`.

## Rationale

We end up setting a value if it is `True`, and otherwise using a default value.

The version with `or` is a little easier to read and avoids duplicating `path_list`.

It works because the left side is evaluated first. If it is evaluated to `True`, `value` will be set to `path_list` and the right side will not be evaluated.
If the result is `False`, the right side will be evaluated and the `value` will be set as the default.

void

## Future Work

void

### DIFF
--- a/src/fprime/fbuild/settings.py
+++ b/src/fprime/fbuild/settings.py
@@ -140,7 +140,7 @@ class IniSettings:
                 settings["settings_file"],
                 key != "install_dest",
             )
-            value = paths_list if paths_list else get_default_value()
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

LeStarch (approved):
Looks good!

PR #104, opened 2022-11-10
Add a `getMaxSize()` Method To Types

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

When looking towards packetized telemetry in F´, we need the ability to determine not just the real size of a type, but the maximum size of that type that is reserved for it in a telemetry packet.

This PR adds those methods, cleans up type importing in the TimeType, and adds UTs.

### DIFF
--- a/src/fprime/common/models/serialize/bool_type.py
+++ b/src/fprime/common/models/serialize/bool_type.py
@@ -44,5 +44,11 @@ class BoolType(ValueType):
         except struct.error:
             raise DeserializeException("Not enough bytes to deserialize bool.")
 
-    def getSize(self):
+    @classmethod
+    def getSize(cls):
         return struct.calcsize("B")
+
+    @classmethod
+    def getMaxSize(cls):
+        """ Maximum size of type """
+        return cls.getSize()  # Always the same as getSize
--- a/src/fprime/common/models/serialize/bool_type.py
+++ b/src/fprime/common/models/serialize/bool_type.py
@@ -50,5 +50,5 @@ class BoolType(ValueType):
 
     @classmethod
     def getMaxSize(cls):
-        """ Maximum size of type """
+        """Maximum size of type"""
         return cls.getSize()  # Always the same as getSize
--- a/src/fprime/common/models/serialize/time_type.py
+++ b/src/fprime/common/models/serialize/time_type.py
@@ -22,7 +22,7 @@ import math
 from enum import Enum
 
 # Custom Python Modules
-import fprime.common.models.serialize.numerical_types
+from fprime.common.models.serialize.numerical_types import U8Type, U16Type, U32Type
 from fprime.common.models.serialize import type_base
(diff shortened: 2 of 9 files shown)


LeStarch:
@thomas-bc @timcanham would y'all take a look?

src/fprime/common/models/serialize/time_type.py
```diff
@@ -209,20 +205,28 @@ def deserialize(self, data, offset):
...
 
         Returns:
             The size of the time type object when serialized
         """
         return (
-            self.__timeBase.getSize()
-            + self.__timeContext.getSize()
-            + self.__secs.getSize()
-            + self.__usecs.getSize()
+            U16Type.getSize() + U8Type.getSize() + U32Type.getSize() + U32Type.getSize()
```
kevin-f-ortega, line 217: I would argue to revert this change. For example, if you change `self.__timeBase`'s type to U8, then this won't work.

src/fprime/common/models/serialize/time_type.py
```diff
@@ -209,20 +205,28 @@ def deserialize(self, data, offset):
...
 
+    @classmethod
+    def getMaxSize(cls):
+        """
+        Return the size of the time type object when serialized
+
+        Returns:
+            The size of the time type object when serialized
+        """
+        return cls.getSize()
```
kevin-f-ortega, line 229: Add the comment `# Always the same as getSize`

src/fprime/common/models/serialize/bool_type.py
```diff
@@ -44,5 +44,11 @@ def deserialize(self, data, offset):
         except struct.error:
             raise DeserializeException("Not enough bytes to deserialize bool.")
 
-    def getSize(self):
+    @classmethod
+    def getSize(cls):
         return struct.calcsize("B")
```
kevin-f-ortega, line 49: I'm just wondering if this `"B"` should be switched to `"?"` throughout this file. I understand both have a standard size of 1 byte, but `"?"` is for boolean, which is the type we are dealing with in this file.

kevin-f-ortega (commented):
Looks good! Just a few comments.

src/fprime/common/models/serialize/time_type.py
LeStarch: Good catch!

LeStarch: This will be fixed when we make the ground system configurable. Until then, this is as it is.

src/fprime/common/models/serialize/bool_type.py
LeStarch: As it turns out we are using a `U8` as our serialize type for booleans. (`0xFF` and `0x00` being the two values).  Thus I'd keep it as-is because the U8 is on-wire type.

kevin-f-ortega (approved):
LGTM

PR #105, opened 2022-11-11
adding default cmake settings

### LINKED ISSUE #1683: Add Default -D flags to `settings.ini`
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

settings.ini should allow a default set of -D flags to submit to a CMake call.

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Enables users to set the `default_cmake_options` `settings.ini` field to pass in default cmake options.

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -401,6 +401,7 @@ class Build:
             default_cmake_args = {
                 option: value
                 for (option, value) in [split_pair(item) for item in default_options]
+                if option != ""
             }
 
             self.cmake.generate_build(
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -311,6 +311,9 @@ class Build:
             if self.get_settings(setting, None) is not None
         }
 
+        # Load in the default settings
+        self.get_settings("default_cmake_options", None)
+
         if "FPRIME_LIBRARY_LOCATIONS" in cmake_args:
             cmake_args["FPRIME_LIBRARY_LOCATIONS"] = ";".join(
                 [str(location) for location in cmake_args["FPRIME_LIBRARY_LOCATIONS"]]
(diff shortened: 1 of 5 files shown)


src/fprime/fbuild/settings.py
```diff
@@ -78,6 +78,7 @@ class IniSettings:
             SettingType.PATH,
             lambda settings: settings["settings_file"],
         ),
+        ("default_cmake_options", SettingType.STRING, ""),
```
kevin-f-ortega, line 81: Format this line like the other entries.

test/fprime/fbuild/test_settings.py
```diff
@@ -39,6 +39,7 @@ def test_settings():
                 "ac_constants": full_path("..") / "config" / "AcConstants.ini",
                 "project_root": full_path(".."),
                 "config_directory": full_path("..") / "config",
+                "default_cmake_options": "",
```
kevin-f-ortega, line 42: Should one of these tests provide a "default_cmake_options" with actual values instead of the empty string?

kevin-f-ortega (commented):
Looks good. Nice work! Only a few comments.

src/fprime/fbuild/settings.py
LeStarch: Cannot do so.  The python formatter has chosen this way.

test/fprime/fbuild/test_settings.py
LeStarch: I will fix this one!  We should test the multi-line setup.

LeStarch: Added a new test case!

LeStarch:
@kevin-f-ortega I added a fix to for the missing test case.  Formatting is at the whim of our formatter.

kevin-f-ortega (approved):
LGTM!

Joshua-Anderson:
Totally late and very optional feedback, but I was wondering if using YAML list instead of newlines to separate options might be a bit cleaner and simpler? I think this is generally a more common YAML practice, I've never actually seen multiline strings without quotes like this before, though it does seem to work fine!

Ex:

```yaml
default_cmake_options:
    - OPTION1=ABC
    - OPTION2=123
    - OPTION3=Something
```

instead of:

```yaml
default_cmake_options: OPTION1=ABC
    OPTION2=123
    OPTION3=Something
```

I figured I might bring this up before this feature goes into regular use and we have to start thinking about backwards compatibility concerns.

src/fprime/fbuild/builder.py
```diff
@@ -311,6 +311,9 @@ def get_cmake_args(self) -> dict:
             if self.get_settings(setting, None) is not None
         }
 
+        # Load in the default settings
+        self.get_settings("default_cmake_options", None)
```
thomas-bc, line 315: @LeStarch this line doesn't seem to do anything, yet it can be misleading to think that it does (as this could potentially tweak the order in which arguments are loaded in). Can you confirm? If so, I'll remove it

thomas-bc: I was looking at this when reviewing https://github.com/fprime-community/fprime-tools/pull/165. Also I don't disagree with Josh's comment above but it may be a little late to switch

PR #107, opened 2022-11-23
Add sanitizer to UT compilation

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**| `fprime-util generate` |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/pull/1772 |

---
## Change Description

Following https://github.com/nasa/fprime/pull/1772, this adds the compiler sanitizers by default on `--ut` targets. An option is added (`fprime-util generate --ut --disable-sanitizers`) in order to disable the sanitizers if needed.

This enables the **leak**, **memory** and **undefined behavior** sanitizers at compilation by passing the associated flags down to CMake. Note that leak is not available on macOS, but this is handled by CMake down at the fprime level (cmake will refuse to activate LSAN on macOS).

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -170,6 +170,7 @@ rstrip
 rsub
 rtd
 rtruediv
+sanitizers
 Sched
 SCLK
 scm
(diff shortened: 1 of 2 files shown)


LeStarch (approved):

LeStarch:
Looks correct to me.  Do you have any test data?  Since it is CLI, I don't expect a UT. If not, I can try it myself.

thomas-bc:
I don't have test cases but I can add that in if necessary. Could be good since it depends on the underlying F' cmake flags not to change. I have tested manually and it's working as expected, I don't think there's any edge case.

PR #111, opened 2022-12-19
Add code formatting command

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**| `fprime-util format` |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/817 |
|**_Has Unit Tests (y/n)_**| N |
|**_Builds Without Errors (y/n)_**| Y |
|**_Documentation Included (y/n)_**| Y |

---
## Change Description

Adds a new command to `fprime-util` to make using clang-format with FPrime source code easier (`fprime-util format`). 

## Rationale

This is mostly needed because of the access specifier macros in FPrime (e.g. `PRIVATE:`), which clang-format can't catch on, which resolves in bad formatting.
This new utility wraps around `clang-format`, replacing the macros before/after processing, and adding some functionality to make it easier to use with FPrime.

Usability is described in `help_text.py` -- can be easily tested within the core F' repo or any deployment, after a `pip install .`
Let me know if you can think of any feature/options that could be useful and I can try to add them in.

### DIFF
--- a/src/fprime/util/cli.py
+++ b/src/fprime/util/cli.py
@@ -154,6 +154,9 @@ def run_code_format(
             file=sys.stderr,
         )
         return 1
+    # Add requested file extensions to the whitelist
+    for file_ext in parsed.whitelist:
+        clang_formatter.whitelist_extension(file_ext)
     # Stage all files that are passed through stdin, if requested
     if parsed.stdin:
         for filename in sys.stdin.read().split():
@@ -243,6 +246,12 @@ def add_special_parsers(
         "--force",
         action="store_true",
         help="Force all listed files to be passed to clang-format (no file extension check)",
+    )
+    format_parser.add_argument(
+        "--whitelist",
+        action="append",
+        default=[],
+        help="Add a file extension to the allowed set",
     )
     format_parser.add_argument(
         "-f",
--- a/src/fprime/util/cli.py
+++ b/src/fprime/util/cli.py
@@ -154,9 +154,9 @@ def run_code_format(
             file=sys.stderr,
         )
         return 1
-    # Add requested file extensions to the whitelist
-    for file_ext in parsed.whitelist:
-        clang_formatter.whitelist_extension(file_ext)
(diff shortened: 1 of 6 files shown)


src/fprime/util/cli.py
```diff
@@ -179,4 +225,54 @@ def add_special_parsers(
...
     )
-    return {"hash-to-file": hash_to_file, "info": print_info, "new": template}
+
+    # Code formatting with clang-format
+    format_parser = subparsers.add_parser(
+        "format",
+        help=help_text.short("format"),
+        description=help_text.long("format"),
+        parents=[common],
+        add_help=True,
```
LeStarch: Typically we don't set the `add_help` flag to `True`.  This is done at the top level.

src/fprime/util/cli.py
```diff
@@ -179,4 +225,54 @@ def add_special_parsers(
...
+        help="Force all listed files to be passed to clang-format (no file extension check)",
+    )
+    format_parser.add_argument(
+        "--whitelist",
+        action="append",
+        default=[],
+        help="Add a file extension to the allowed set",
+    )
+    format_parser.add_argument(
+        "-f",
```
LeStarch, line 257: Should we make these positional inputs?  It would be more intuitive to users to not need to specify -f just to format.

LeStarch (changes requested):

src/fprime/util/cli.py
thomas-bc: Changed back to False

src/fprime/util/cli.py
thomas-bc: I agree that it would be more intuitive. However because the common parser has the `platform` optional positional argument, I believe this is not possible since the parser can't tell whether the first positional is `platform` or `files`. Unless I missed an option to influence the parsing or remove the platform argument from the child parser.

LeStarch (approved):

PR #117, opened 2023-02-23
Improve FppMissingSupportFiles error message

### LINKED ISSUE #1888: Improve Error Message with FPP Check
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

See error message discussion in #1801

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/1888 |

---
## Change Description

Fixes https://github.com/nasa/fprime/issues/1888. 

I wasn't sure about the wording, this is the best that I could come up with in a concise manner. Let me know if you can think of better.

### DIFF
--- a/src/fprime/fpp/common.py
+++ b/src/fprime/fpp/common.py
@@ -19,7 +19,7 @@ class FppMissingSupportFiles(FprimeException):
 
     def __init__(self, file):
         super().__init__(
-            f"{file} not found. Ensure that you are running from the correct directory and using fprime v3.1.0 or higher."
+            f"Current directory does not define any FPP files. Did you intend to run in the topology directory?"
         )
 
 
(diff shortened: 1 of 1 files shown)


LeStarch:
It seems like the commands should work independently of current working directory.  Do they not do this?

thomas-bc:
The command is looking for the fpp-input-list under the build cache in the same `context` as where it is ran from. So if the current directory does not define FPP files, fpp-input-list is not generated by CMake in that context, and the command errors.
https://github.com/fprime-community/fprime-tools/blob/0c668dc8145ef5b806f6f892723ca5694f00d58a/src/fprime/fpp/common.py#L78-L81

LeStarch (approved):

PR #119, opened 2023-02-27
Add Python 3.11 to integration tests

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Adding Python 3.11 to the integration test workflow. 
I also set the `strategy.fail-fast` to false. Default was True which caused all integration tests to abort if one failed, which can give a wrong impression of what's happening and take longer to review. That way, we know which ones are failing and which ones are not, without having to dig through the logs.
I also removed some unused variables that were probably copy/pasted from somewhere else, as it doesn't look like they were ever used.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -22,7 +22,7 @@ jobs:
           python -m pip install --upgrade pip urllib3 setuptools_scm
       - name: clone fprime and gps-tutorial
         run: |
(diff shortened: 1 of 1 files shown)


LeStarch (approved):
Note, we should add F´ 3.1.1 to this list at some point too.

PR #122, opened 2023-03-29
fixing gcovr usage for -p and -d flags

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Changes gcov handling:
1. Improves autocoding to allow including via flags: --all-ac, --comp-ac, --port-ac, --type-ac to include each of the types of Ac
2. Removes the special ac directory called `{{AUTOCODE}}`. Modern gcovr doesn't support it and it was a maintenance nightmare
3. Fixed handling of `-d` and `-p` flags

### DIFF
--- a/src/fprime/fbuild/gcovr.py
+++ b/src/fprime/fbuild/gcovr.py
@@ -1,5 +1,6 @@
 """ fprime.fbuild.gcovr: coverage target implementations """
 import atexit
+import itertools
 import shutil
 import subprocess
 import sys
@@ -7,8 +8,6 @@ from pathlib import Path
 from typing import Dict, List, Tuple, Union
 
 from .target import ExecutableAction, Target, TargetScope, CompositeTarget
-
-TEMPORARY_DIRECTORY = "{{AUTOCODE}}"
 
 
 def _get_project_path(builder: "Build", module: Union[str, Path]) -> Path:
@@ -36,24 +35,6 @@ def _using_root(builder: "Build", context: Path, scope: TargetScope):
     return builder.is_deployment(context) or scope == TargetScope.GLOBAL
 
 
-def _get_ac_directory(builder: "Build", context: Path, scope: TargetScope) -> Path:
-    """Gets a directory for AC files to be placed
-
-    AC files need to migrate out of the build cache and into the build for purposes of HTML coverage.  This selects an
-    appropriate location for this temporary directory.
-
-    Args:
-        builder: build to use
-        context: path context to use
-    """
-    base_path = (
-        _get_project_path(builder, ".")
-        if _using_root(builder, context, scope)
-        else context
-    )
-    return (base_path / TEMPORARY_DIRECTORY).absolute()
-
-
 class GcovClean(ExecutableAction):
     """Target used to scrub gcov files before a coverage run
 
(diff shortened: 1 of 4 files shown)


src/fprime/fbuild/gcovr.py
```diff
@@ -227,60 +103,108 @@ def execute(
...
         filter_path = (
-            Path(project_root).absolute()
-            if _using_root(builder, context, self.scope)
-            else _get_project_path(builder, context)
+            (
+                Path(project_root).absolute()
+                if _using_root(builder, context, self.scope)
+                else _get_project_path(builder, context)
+            )
+            .absolute()
```
thomas-bc, line 155: From [the docs on resolve()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve) : 
> **Path.resolve:** Make the path absolute, resolving any symlinks

So .absolute() would be redundant here? And it seems like there were issues/debates around .absolute() prior to Python 3.11 (see [here](https://bugs.python.org/issue29688)) so it might be wise to get rid of it.

thomas-bc: From [the docs on resolve()](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve) :

> **Path.resolve:** Make the path absolute, resolving any symlinks

So .absolute() would be redundant here? And it seems like there were issues/debates around .absolute() prior to Python 3.11 (see [here](https://bugs.python.org/issue29688)) so it might be wise to get rid of it.

src/fprime/util/build_helper.py
```diff
@@ -269,5 +269,6 @@ def utility_entry(args):
         return 1
     except Exception as exc:
         print(f"[ERROR] {exc}", file=sys.stderr)
+        raise
         return 1
```
thomas-bc: This is now unreached

src/fprime/fbuild/target.py
```diff
@@ -56,10 +57,17 @@ def is_supported(self, builder: "Build", context: Path):
...
-        self, builder: "Build", context: Path, args: Tuple[Dict[str, str], List[str]]
+        self,
+        builder: "Build",
+        context: Path,
+        args: Tuple[Dict[str, str], List[str], Dict[str, bool]],
     ):
         """Executes the given target"""
 
+    def option_args(self):
+        """List of option arguments handled by this target"""
```
thomas-bc: It is kind of implied, but it could be good to explicitly mention here that options are default to False, for future developers/targets that may want to use this.

``` 
option_args = {
            flag: getattr(parsed, flag.lstrip("-").replace("-", "_"), False)
[...]
```

thomas-bc (commented):
I added a couple of comments but this looks good!

src/fprime/fbuild/gcovr.py
LeStarch: will switch to `.resolve()`

src/fprime/util/build_helper.py
LeStarch: For debugging, will revert.

src/fprime/fbuild/target.py
LeStarch: Agreed.  I will extend the comment

LeStarch:
Ready for re-review.

thomas-bc (approved):

PR #125, opened 2023-04-14
New Component (FPP) generation feature

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Affected Architectures(s)_**| `fprime-util new` |
|**_Related Issue(s)_**| https://github.com/fprime-community/fprime-tools/issues/37   https://github.com/nasa/fprime/issues/1961 |
|**_Builds Without Errors (y/n)_**| y |

---
## Change Description

`fprime-util new --component` changed to generate FPP instead of XML.

## Future work

Ports are much much easier to create in FPP than in XML and probably do not require a `new --port` functionality. Should we get rid of it? Michael you mentioned a `new --module` idea instead.


src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/sdd.md
```diff
@@ -31,9 +29,25 @@ Add component states in the chart below
 Add sequence diagrams here
 
 ## Parameters
+| Name | Description |
```
LeStarch, line 32: Should these be driven by the choices made?

LeStarch: At some point I want to auto-generate these sections of the SDD....so maybe it doesn't matter.

src/fprime/util/cookiecutter_wrapper.py
```diff
@@ -77,23 +55,22 @@ def run_impl(deployment: Path, path: Path, platform: str, verbose: bool):
...
+    shutil.move(hpp_src, hpp_dest)
+    shutil.move(cpp_src, cpp_dest)
+
     return True
 
 
-def add_to_cmake(list_file: Path, comp_path: Path):
-    """Adds new component or port to CMakeLists.txt"""
-    print(f"[INFO] Found CMakeLists.txt at '{list_file}'")
+def add_to_cmake(list_file: Path, comp_path: Path, project_root: Path = None):
```
LeStarch, line 81: This function doesn't look like it will work with arbitrary CMake structure.

e.g. it only works with the parent cmake list, not `project.cmake` nor if there are empty directories in between.

In other implementations of this we have recursed upward to project root looking for:

1. If `project.cmake` then add to it
2. Else if `CMakeLists.txt` then add to it
3. Else, repeat on parent directory

src/fprime/util/cookiecutter_wrapper.py
```diff
@@ -77,23 +55,22 @@ def run_impl(deployment: Path, path: Path, platform: str, verbose: bool):
...
-            lines = file_handle.readlines()
-        lines = lines[11:]  # Remove old header
-        with open(dest, "a") as file_handle:
-            file_handle.write("".join(lines))
-
-    removals = [Path(hpp_src), Path(cpp_src)]
-    for removal in removals:
-        os.remove(removal)
+    # Move (and overwrite) files from *.(c|h)pp-template to *.(c|h)pp
+    shutil.move(hpp_src, hpp_dest)
```
LeStarch, line 60: Was expecting format to run here.

LeStarch (changes requested):

src/fprime/util/cookiecutter_wrapper.py
thomas-bc: Added

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/sdd.md
thomas-bc: We can add that really easily, my thought process on not doing it is that the presence of the information "there is nothing" has a stronger meaning than the lack of information. So the presence of an empty _Parameters_ table is more information than the lack of a _Parameter_ table. But that's def in the personal-preference space... I might be looking at it wrong

src/fprime/util/cookiecutter_wrapper.py
thomas-bc: Added my take at implementing this, let me know how that looks

src/fprime/cookiecutter_templates/cookiecutter-fprime-component/{{cookiecutter.component_name}}/docs/sdd.md
LeStarch: Agreed.

LeStarch (approved):

PR #131, opened 2023-04-23
Fixing python executable when only python3 exists

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

`fprime-util new --project` calls `python` directly and does not use the `sys.executable` reference to the executable.  Thus on systems where `python3` exists but not `python` the above call will break.

This PR switches that use to `sys.executable` to ensure a valid python will always be used.

### DIFF
--- a/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/hooks/post_gen_project.py
+++ b/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/hooks/post_gen_project.py
@@ -9,7 +9,7 @@ It does the following:
 
 @author thomas-bc
 """
-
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):
Ah yes, good point. Thanks !

LeStarch:
@thomas-bc when you see this, could you review.  If you agree with it, merge it, and make a hotfix release (`v3.2.1`) as we need it next week.

LeStarch:
@thomas-bc never-mind I see your approval, I'll do the rest.  Didn't expect a response this late.

PR #136, opened 2023-05-02
Fix #130

### LINKED ISSUE #130: (F841) Local variable name is assigned to but never used
| | |
|:---|:---|
|**_F´ Version_**| devel |
|**_Affected Component_**| string_util.py  |
---
## Problem Description

Ruff linter spotted one issue in:

https://github.com/fprime-community/fprime-tools/blob/58e28dfa4a1c93e55aff533747def2b5386aa40f/src/fprime/util/string_util.py#L95

```
src/fprime/util/string_util.py:95:26: F841 [*] Local variable `e` is assigned to but never used
```

According to Flake 8: "A local variable in your function was defined but is not being used. This local variable should be removed."

Ref: https://www.flake8rules.com/rules/F841.html

## Expected Behavior

The variable should be used or be removed.

PS: can be tagged as `Good First Issue`

stepanzubkov:
Just remove unused variable

### DIFF
--- a/src/fprime/util/string_util.py
+++ b/src/fprime/util/string_util.py
@@ -92,7 +92,7 @@ def format_string_template(format_str, given_values):
         result = formatted_str.format(*values)
         result = result.replace("%%", "%")
         return result
-    except ValueError as e:
+    except ValueError:
         msg = "Value and format string do not match. "
         msg += " Will ignore integer flags `d` in string template. "
         msg += f"values: {values}. "


ThibFrgsGmz:
You can put in the title the hyperlink to the issue, it will render the same but with a link to directly access the ticket 😀

src/fprime/util/string_util.py
```diff
@@ -92,7 +92,7 @@ def convert_ignore_int(match_obj):
         result = formatted_str.format(*values)
         result = result.replace("%%", "%")
         return result
-    except ValueError as e:
+    except ValueError:
```
thomas-bc, line 95: It seems there is this exact pattern that's used [a couple lines below](https://github.com/fprime-community/fprime-tools/blob/53ccc63bc241c3a902e7d5c114fcf85d5ae178cf/src/fprime/util/string_util.py#L110), we should probably keep it consistent. Would you mind making that change as well?

thomas-bc (changes requested):

src/fprime/util/string_util.py
thomas-bc: Oh wait I may have spoken too fast, we're raising it up in that second case.
https://github.com/fprime-community/fprime-tools/blob/6d258007391e9a4091b29b0af3ba6ef13c798f5b/src/fprime/util/string_util.py#L116

thomas-bc: Annnd we're actually using the exception...
https://github.com/fprime-community/fprime-tools/blob/6d258007391e9a4091b29b0af3ba6ef13c798f5b/src/fprime/util/string_util.py#L114
Read that way too fast lol, my apologies. This looks good to me.

thomas-bc (approved):
LGTM

ThibFrgsGmz:
The title update did not worked ? 🤔
At least in the description it would work.

PR #138, opened 2023-05-12
Add `flake8-error-message` convention

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This feature implements the [flake8-err-msg](https://pypi.org/project/flake8-errmsg/) which allows the formatting of nice error messages.

## Rationale

The problem is that Python includes the line with the lift in the default traceback. This means that a user receives a message like this:

```py
sub = "Some value"
raise RuntimeError(f"{sub!r} is incorrect")
```

```sh
Traceback (most recent call last):
  File "tmp.py", line 2, in <module>
    raise RuntimeError(f"{sub!r} is incorrect")
RuntimeError: 'Some value' is incorrect
```

If it is longer or more complex, the duplication can be confusing for a user not used to reading tracebacks.

On the other hand, if you always assign to something like msg, you get :

```py
sub = "Some value"
msg = f"{sub!r} is incorrect"
raise RunetimeError(msg)
```

```sh
Traceback (most recent call last):
  File "tmp.py", line 3, in <module>
    raise RuntimeError(msg)
RuntimeError: 'Some value' is incorrect
```

There is now a more straightforward trace, less code and no double messages.

[Ref - Ruff linter](https://beta.ruff.rs/docs/rules/f-string-in-exception/)

Inspection is enough

## Future Work

https://github.com/fprime-community/fprime-tools/pull/140

### DIFF
--- a/src/fprime/common/models/serialize/enum_type.py
+++ b/src/fprime/common/models/serialize/enum_type.py
@@ -77,9 +77,7 @@ class EnumType(DictionaryType):
             int_val = struct.unpack_from(">i", data, offset)[0]
         except struct.error:
             msg = f"Could not deserialize enum value. Needed: {self.getSize()} bytes Found: {len(data[offset:])}"
-            raise DeserializeException(
-                msg
-            )
+            raise DeserializeException(msg)
         for key, val in self.ENUM_DICT.items():
             if int_val == val:
                 self._val = key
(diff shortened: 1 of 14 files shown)


thomas-bc:
See discussion in https://github.com/fprime-community/fprime-gds/pull/129

thomas-bc (approved):

LeStarch (approved):

LeStarch:
@thomas-bc merge when you are ready!

LeStarch:
We agreed that these changes are good, however, as a note: in F´  we expect to trap exceptions and present a clean error message to users without a confusing stack-trace. Thus the motivation here becomes one of appeasing a potentially good linter and less to beautify messages for users directly.

PR #139, opened 2023-05-12
Add consistency to imports based on `isort` settings

ThibFrgsGmz:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

This PR consists of de-duplicating, grouping and sorting imports according to isort parameters.

## Rationale

This is to make the code base more consistent, as using a common convention for imports makes the code more readable and idiomatic.

[Ref - Ruff linter](https://beta.ruff.rs/docs/rules/unsorted-imports/)

Visual inspection

## Future Work

Add Ruff linter into repo's CI (https://github.com/fprime-community/fprime-tools/pull/140).

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -7,21 +7,20 @@ import re
 from pathlib import Path
 from typing import Iterable, List, Union
 
+# Forces targets into existence
+import fprime.fbuild.target_definitions  # lgtm[py/unused-import]
+from fprime.common.error import FprimeException
+from fprime.fbuild.cmake import CMakeException, CMakeHandler
+from fprime.fbuild.settings import IniSettings
+from fprime.fbuild.target import Target, TargetScope
 from fprime.fbuild.types import (
+    AmbiguousToolchainException,
     BuildType,
     InvalidBuildCacheException,
     MissingBuildCachePath,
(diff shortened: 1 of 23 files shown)


src/fprime/fbuild/builder.py
```diff
@@ -7,21 +7,20 @@
 from pathlib import Path
 from typing import Iterable, List, Union
 
+# Forces targets into existence
+import fprime.fbuild.target_definitions  # lgtm[py/unused-import]
```
thomas-bc, line 11: This import here is quite special and I wonder if we should leave it at the very bottom of the imports, to make it obvious/easily identifiable.

thomas-bc (commented):
LGTM. Added a comment for your consideration.

src/fprime/fbuild/builder.py
thomas-bc: Also, we should probably simply revise that pattern.

LeStarch: We agreed this was ok, but revision of the pattern is advisable.

LeStarch (approved):

LeStarch:
@thomas-bc looks good to me.  All yours to close-out.

PR #141, opened 2023-05-24
Fix urllib3 version to prevent breakages on macOS

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

urllib3>=2.0.0 breaks on macOS over SSL versioning.  This fixes the package version until `urllib3` comes up with a solution.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ to interact with the data coming from the FSW.
         "Cheetah3>=3.2.6",
         "cookiecutter>=1.7.2",
         "gcovr>=6.0",
-        "urllib3==1.26.16",  # Cannot upgrade to 2.0.0 until SSL works on macOS
+        "urllib3<2.0.0",
     ],
     extras_require={
         "dev": [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -88,6 +88,7 @@
         "Cheetah3>=3.2.6",
         "cookiecutter>=1.7.2",
         "gcovr>=6.0",
+        "urllib3==1.26.16",  # Cannot upgrade to 2.0.0 until SSL works on macOS
```
thomas-bc: We should probably do `urllib3<2.0.0` instead, letting pip do its thing, and also mostly just in case the SSL fix doesn't come out for a while and `1.26.x` keeps being supported/fixed on its own?

thomas-bc (commented):
Just clarifying something, it breaks on macOS' system Python, not all macOS installs, right?
Reference, for posterity: https://github.com/urllib3/urllib3/issues/3020

LeStarch:
@thomas-bc yes this only applies to the system python on macOS, which is specifically recommended against using (it is for OS support)....however, even seasoned mac users don't know this and I don't want to have to document this "correct fix".

thomas-bc (approved):

PR #142, opened 2023-05-30
Refactor `new` build system structure (nasa#1994)

thomas-bc:
| | |
|:---|:---|
|**_Affected Architectures(s)_**| `fprime-util new` |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/pull/1994 https://github.com/nasa/fprime/issues/2002 |

---
## Change Description

Addresses https://github.com/nasa/fprime/issues/2002

https://github.com/nasa/fprime/pull/1994 introduced a new structure for projects where the CMake project() can be defined at the top level of the project tree only, and not within deployments. This modifies the `fprime-util new` commands to account for that.

I also took the liberty to add the `Components/` module to the project generation tree because it sounds to me like a good pattern to suggest to users - while it does not force anything onto them.

**Note:** I had to revert back the default fprime version `new --project` pulls since #1994 is only in devel for now.

Checkout, install, and verify that `fprime-util new --project && fprime-util new --deployment` generates a buildable and runnable project.

## Future Work

- Switch default project F´ version back to master for next release.
- _Somewhat unrelated_, `fprime-gds` should automatically detect the deployment's `build-artifacts/` when ran within the deployment module. Currently, we need to pass in `-d ../build-artifacts/MyDeployment` or run at the top level of the project tree.

### DIFF
--- a/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/cookiecutter.json
+++ b/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/cookiecutter.json
@@ -1,6 +1,7 @@
 {
+    "__default_branch": "master",
     "project_name": "MyProject",
-    "fprime_branch_or_tag": "devel",
+    "fprime_branch_or_tag": "{{ cookiecutter.__default_branch }}",
     "install_venv": ["yes", "no"],
     "venv_install_path": "{% if cookiecutter.install_venv == 'yes' %}./venv{% else %}None{% endif %}"
 }
--- a/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/cookiecutter.json
+++ b/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/cookiecutter.json
@@ -1,5 +1,5 @@
 {
-    "__default_branch": "master",
+    "__default_branch": "devel",
     "project_name": "MyProject",
     "fprime_branch_or_tag": "{{ cookiecutter.__default_branch }}",
     "install_venv": ["yes", "no"],
--- a/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/cookiecutter.json
+++ b/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/cookiecutter.json
@@ -1,6 +1,6 @@
 {
     "project_name": "MyProject",
-    "fprime_branch_or_tag": "master",
+    "fprime_branch_or_tag": "devel",
     "install_venv": ["yes", "no"],
     "venv_install_path": "{% if cookiecutter.install_venv == 'yes' %}./venv{% else %}None{% endif %}"
 }
--- a/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/hooks/post_gen_project.py
+++ b/src/fprime/cookiecutter_templates/cookiecutter-fprime-project/hooks/post_gen_project.py
@@ -12,7 +12,8 @@ It does the following:
 import subprocess
 import sys
 
-DEFAULT_BRANCH = "devel"
+# Fail-safe in case user input is invalid branch/tag
+DEFAULT_BRANCH = "{{ cookiecutter.__default_branch }}"
 
 # Add F' as a submodule
 subprocess.run(["git", "init"])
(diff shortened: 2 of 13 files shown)


src/fprime/cookiecutter_templates/cookiecutter-fprime-project/cookiecutter.json
```diff
@@ -1,6 +1,6 @@
 {
     "project_name": "MyProject",
-    "fprime_branch_or_tag": "master",
+    "fprime_branch_or_tag": "devel",
```
LeStarch: Why?  I thought we wanted master.

src/fprime/cookiecutter_templates/cookiecutter-fprime-project/hooks/post_gen_project.py
```diff
@@ -12,7 +12,7 @@
 import subprocess
 import sys
 
-DEFAULT_BRANCH = "master"
+DEFAULT_BRANCH = "devel"
```
LeStarch: Is it possible to remove this defined-in-multiple-places for this tag?

src/fprime/cookiecutter_templates/cookiecutter-fprime-project/{{cookiecutter.project_name}}/project.cmake
```diff
@@ -1,2 +1,4 @@
-# This CMake file is intended to register project-wide objects so they can be
-# reused easily between deployments, but also by other projects.
+# This CMake file is intended to register project-wide objects.
+# This allows for reuse between deployments, or other projects.
+
+add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/Components")
```
LeStarch, line 4: This seems to require `Components` as opposed to just using it when the user creates this.  Intended?

LeStarch (commented):
A few minor questions, otherwise looks good!

src/fprime/cookiecutter_templates/cookiecutter-fprime-project/cookiecutter.json
thomas-bc: nasa/fprime#1994 is only in devel for now, so pulling master would end up creating a `new --deployment` that would error on generation

thomas-bc: I noted switching back to master as Future Work for next release

src/fprime/cookiecutter_templates/cookiecutter-fprime-project/{{cookiecutter.project_name}}/project.cmake
thomas-bc: Yes, I have added the `Components/` module as part of the project generation. I can revert back, but ...

>I also took the liberty to add the Components/ module to the project generation tree because it sounds to me like a good pattern to suggest to users - while it does not force anything onto them.

src/fprime/cookiecutter_templates/cookiecutter-fprime-project/cookiecutter.json
thomas-bc: Oh wait - I thought the change I made was from devel to master. Lol. Will revert back.

src/fprime/cookiecutter_templates/cookiecutter-fprime-project/hooks/post_gen_project.py
thomas-bc: Yeah it will be a little quirky cause we don't simply want to pull out `fprime_branch_or_tag` after the user has set it - as this `DEFAULT_BRANCH` really is a failsafe in case the user input is not a valid branch/tag. And cookiecutter doesn't allow to ingest/push that value into the JSON I believe.
Implementing the less quirky/most verbose way I can think of.

thomas-bc: Pushed - let me know what your think

LeStarch: Great!

src/fprime/cookiecutter_templates/cookiecutter-fprime-project/{{cookiecutter.project_name}}/project.cmake
LeStarch: It is fine!

LeStarch (approved):

PR #147, opened 2023-06-16
Feature: FPP Visualize

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2059 |

---
## Change Description

**Depends on**:
- https://github.com/fprime-community/fprime-visual/pull/1
- https://github.com/nasa/fprime/issues/2084

Introducing: `fprime-util visualize`
Within a Topology directory, performs the following

-  fpp-to-xml
-  fpl-convert-xml
-  fpl-layout
-  start fprime-visual Flask app to serve visualization

Effectively, this allows with a single command to spin up a visualization of the topology and all its subtopology graphs.

```bash
cd Ref/Top
fprime-util visualize
```

## Rationale

Major QoL improvement to visualization capabilities. This is currently doable, but requires a lot of manual installation of tools etc...

## Todo
- The 2 things listed in _Depends on_
- Probably add some tests

## Note

Opening as draft in case there's change requests. Do not merge for now, there are a couple of dependencies for this that need to be released before this can go through.

### DIFF
--- a/src/fprime/fpp/cli.py
+++ b/src/fprime/fpp/cli.py
@@ -34,6 +34,30 @@ def run_fpp_check(
         args=({}, ["-u", parsed.unconnected] if parsed.unconnected else []),
     )
 
+def run_fpp_to_xml(
+    build: "Build",
+    parsed: argparse.Namespace,
+    _: Dict[str, str],
+    __: Dict[str, str],
+    ___: List[str],
+):
+    """Run fpp check application
+
+    Handles the fpp-check endpoint by running the utility fpp-check.
+
+    Args:
+        build: build directory output
+        parsed: parsed input arguments
+        _: unused cmake_args
+        __: unused make_args
+        ___: unused pass-through arguments
+    """
+    FppUtility("fpp-to-xml").execute(
+        build,
+        parsed.path,
+        args=({}, ["--directory", parsed.directory] if parsed.directory else []),
+    )
+
 
 def add_fpp_parsers(
     subparsers, common: argparse.ArgumentParser
@@ -56,4 +80,10 @@ def add_fpp_parsers(
     check_parser.add_argument(
         "-u", "--unconnected", default=None, help="write unconnected ports to file"
     )
-    return {"fpp-check": run_fpp_check}, {"fpp-check": check_parser}
+    fpp_to_xml_parser = subparsers.add_parser(
+        "fpp-to-xml", help="Runs fpp-to-xml utility", parents=[common], add_help=False
+    )
+    fpp_to_xml_parser.add_argument(
+        "--directory", default=None, help="Output directory"
+    )
+    return {"fpp-check": run_fpp_check, "fpp-to-xml": run_fpp_to_xml}, {"fpp-check": check_parser, "fpp-to-xml": fpp_to_xml_parser}
--- a/src/fprime/fpp/cli.py
+++ b/src/fprime/fpp/cli.py
@@ -34,6 +34,7 @@ def run_fpp_check(
         args=({}, ["-u", parsed.unconnected] if parsed.unconnected else []),
     )
 
+
 def run_fpp_to_xml(
     build: "Build",
     parsed: argparse.Namespace,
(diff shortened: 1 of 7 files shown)


thomas-bc:
Fails tests because `fprime_visual` pip package is not currently published.

src/fprime/fpp/visualize.py
```diff
@@ -0,0 +1,136 @@
...
+
+    Args:
+        build: build directory output
+        parsed: parsed input arguments
+        _: unused cmake_args
+        __: unused make_args
+        ___: unused pass-through arguments
+    """
+
+    # We could use the build directory, but it's quirky because not managed by CMake ??
```
thomas-bc: Thougts?

src/fprime/fpp/visualize.py
```diff
@@ -0,0 +1,136 @@
...
+    xml_cache = Path(".fpp-viz-cache/xml")
+    xml_cache.mkdir(parents=True, exist_ok=True)
+
+    # Run fpp-to-xml
+    FppUtility("fpp-to-xml").execute(
+        build,
+        parsed.path,
+        args=(
+            {},
+            ["--directory", ".fpp-viz-cache/xml"],
```
thomas-bc: Need rework so `--directory` influences the top level `.fpp-viz-cache` only and everything flows down from it.

src/fprime/fpp/visualize.py
thomas-bc: Using `.visualize-cache/` in current directory for now, let me know if you think we should change that.

thomas-bc:
Ready for review, as all the pieces are there, see https://github.com/nasa/fprime/issues/2059#issuecomment-1622704571

src/fprime/fpp/cli.py
```diff
@@ -35,6 +35,31 @@ def run_fpp_check(
...
 
 
+def run_fpp_to_xml(
+    build: "Build",
+    parsed: argparse.Namespace,
+    _: Dict[str, str],
+    __: Dict[str, str],
+    ___: List[str],
+):
+    """Run fpp check application
```
LeStarch: Comment out of date.  Should this not say something about fpp-to-xml.

src/fprime/fpp/cli.py
```diff
@@ -56,4 +81,11 @@ def add_fpp_parsers(
     check_parser.add_argument(
         "-u", "--unconnected", default=None, help="write unconnected ports to file"
     )
-    return {"fpp-check": run_fpp_check}, {"fpp-check": check_parser}
+    fpp_to_xml_parser = subparsers.add_parser(
+        "fpp-to-xml", help="Runs fpp-to-xml utility", parents=[common], add_help=False
```
LeStarch, line 83: Does `fpp-to-xml` need to be a public thing?  Could it not just be an internal function?  Given that we intend to remove XML it seems like maybe we should just internalize this.

src/fprime/fpp/visualize.py
```diff
@@ -0,0 +1,154 @@
...
+        "--gui-port",
+        help="Set the GUI port for fprime-visual [default: %(default)s]",
+        required=False,
+        default=7000,
+    )
+    viz_parser.add_argument(
+        "--cache-dir",
+        help="Set the directory to store layout files in [default: %(default)s]",
+        required=False,
+        default=".visualize-cache",
```
LeStarch: Should the default be in `/tmp`?  I don't see reason to keep it around locally.

LeStarch (commented):
A few questions, but looks generally good.

src/fprime/fpp/visualize.py
thomas-bc: True - if we expose the option to change where it goes, might as well make the default `/tmp`. Will change

src/fprime/fpp/cli.py
thomas-bc: Discussed internally, we will keep fpp-to-xml public for now

src/fprime/fpp/visualize.py
thomas-bc: Changed to `/tmp/fprime-visualize`

src/fprime/fpp/visualize.py
```diff
@@ -146,9 +146,9 @@ def add_fpp_viz_parsers(
         default=7000,
     )
     viz_parser.add_argument(
-        "--cache-dir",
+        "--working-dir",
         help="Set the directory to store layout files in [default: %(default)s]",
         required=False,
-        default=".visualize-cache",
+        default="/tmp/fprime-visualize",
```
LeStarch: What happens when this directory does not give write permissions?  Do we tell the user that they should override with the `--working-dir` flag.

LeStarch (dismissed):
I had one more minor question, but not sure it should trip up the PR.

src/fprime/fpp/visualize.py
thomas-bc: Good point, I added a more verbose error with resolution approach.

LeStarch (approved):

PR #154, opened 2023-07-25
Validate for new build structure

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2094 |

---
## Change Description

Making sure we're still relevant for the new build structure introduced by https://github.com/nasa/fprime/pull/1994
Correcting "assumptions" in the code that _deployment==cmake project_, which mostly came down to variable names and vocabulary.
I had to scrub a minor thing in `new --component` that would use the deployment's name as default prompt for component namespace, if the command is ran from a deployment. I thought about switching to using CWD's name, since that's usually what I've seen being used.

I tested as much as I could with projects on the old and new build structure, as well as with core fprime.

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -442,7 +442,7 @@ class Build:
         exist.
 
         Notes:
-            Historically, the root of the CMake project was an F' deployment. In F' > v3.2.0, this is more often an 
+            Historically, the root of the CMake project was an F' deployment. In F' > v3.2.0, this is more often an
             F' project root.
 
         Returns;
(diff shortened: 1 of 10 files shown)


LeStarch:
"I thought about switching to using CWD's name, since that's usually what I've seen being used."  Yes! Do this!

src/fprime/fbuild/builder.py
```diff
@@ -41,27 +41,27 @@ class Build:
...
         """Constructs a build object from its constituent parts
 
         Args:
             build_type: member of the enum BuildType specifying fprime build type
-            deployment: path to deployment that this build represents
+            project: path to cmake project that this build represents
         """
         self.build_type = build_type
-        self.deployment = deployment
+        self.cmake_root = project
```
LeStarch, line 64: Is this  a default, or is it always like this?

LeStarch (dismissed):
Looks good!

LeStarch (approved):

PR #165, opened 2023-09-07
Fix arguments order for generate stage

SMorettini:
| | |
|:---|:---|
|**_Affected Component_**| fprime-util generate  |

---
## Change Description

Cmake arguments come from three different places: 
1. default
2. settings.ini
3. user-defined arguments when running fprime-util generate

Before this change, the order was the default, user-defined and settings.ini. Now I changed the order as reported above.

## Rationale

I was trying to set the -DCMAKE_BUILD_TYPE=Release  during the generation of the unit tests but I was not able to send anything else than Debug. This was happening because when the arguments were read from  `settings.ini`, the build type was put by default to Debug. This overrides the build type set from the command line. 

With my change, the user-defined variable when running `fprime-util generate ...` overrides any other variable.

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -329,7 +329,9 @@ class Build:
             and self.build_type == BuildType.BUILD_TESTING
         ):
             cmake_args["BUILD_TESTING"] = "ON"
-            cmake_args["CMAKE_BUILD_TYPE"] = user_cmake_args.get("CMAKE_BUILD_TYPE", "Debug")
+            cmake_args["CMAKE_BUILD_TYPE"] = user_cmake_args.get(
+                "CMAKE_BUILD_TYPE", "Debug"
+            )
         elif self.build_type == BuildType.BUILD_TESTING:
             cmake_args["CMAKE_BUILD_TYPE"] = "Testing"
         return cmake_args
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -396,11 +396,20 @@ class Build:
             cmake_args.update(user_cmake_args)  # User-supplied values from command line
             cmake_args.update(self.get_cmake_args())  # FPRIME_* values (settings.ini)
 
-            if self.build_type == BuildType.BUILD_TESTING:
+            # When the new v3 autocoder directory exists, this means we can use the new UT api and preserve the build type
+            v3_autocoder_directory = Path(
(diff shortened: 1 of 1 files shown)


thomas-bc (dismissed):
LGTM! Agreed that CLI args should take precedence.

LeStarch (approved):

thomas-bc:
@SMorettini turns out the order of dictionary unpacking was a conscious decision, as users should not override the `FPRIME_*` variables from get_cmake_args() through the command line, but instead use the settings.ini file. Otherwise issues can arise. This means `CMAKE_BUILD_TYPE` should not be set as part of the get_cmake_args() call, but rather at the generate() level. 

I went ahead and implemented this in your PR branch, I hope you don't mind.  Would you want to give this a quick review? 

I tested locally and it should behave as you requested. Would very much appreciate some additional eyes on this though!
Thanks!

thomas-bc:
Merging, let me know if you have any comments!

src/fprime/fbuild/builder.py
```diff
@@ -399,16 +386,35 @@ def split_pair(item):
...
+                cmake_args.get("FPRIME_FRAMEWORK_PATH") / "cmake" / "autocoder"
+            )
+            if (
+                v3_autocoder_directory.exists()
+                and self.build_type == BuildType.BUILD_TESTING
+            ):
+                cmake_args["BUILD_TESTING"] = "ON"
+                cmake_args["CMAKE_BUILD_TYPE"] = user_cmake_args.get(
+                    "CMAKE_BUILD_TYPE", "Debug"
+                )
```
SMorettini, line 410: @thomas-bc  I originally opened this pull request to give the possibility to build UTs using a `CMAKE_BUILD_TYPE` different from `Debug`. I agree with your changes but if I understand correctly the code, you reintroduced back the problem I wanted to remove. 

Am I right or the `CMAKE_BUILD_TYPE` will be always `Debug` if the UTs are built? Is that something that we want by design? I think that the user should have the flexibility to choose the `CMAKE_BUILD_TYPE` in any case. 
If by design we don't want to give this flexibility then we should document it somewhere because, at least for me, it seems a hidden and unexpected feature.

SMorettini: @thomas-bc do you have any comments about this?

LeStarch: For newer versions of F´ (v3+) it only defaults to debug if the user has not set "CMAKE_BUILD_TYPE"

LeStarch: I tested this and a custom build type does carry into a UT cache!

PR #171, opened 2023-10-11
Reimplement `impl` with fpp-to-cpp

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Implements https://github.com/nasa/fprime/issues/2204

`fprime-util impl [--ut]` now uses fpp-to-cpp instead of the Python autocoders (through cmake)
Also uses the same interface to run the implementation after `fprime-util new --component`

Also some small changes to help text here and there.

### DIFF
--- a/src/fprime/fpp/common.py
+++ b/src/fprime/fpp/common.py
@@ -132,7 +132,7 @@ class FppUtility(ExecutableAction):
         imports, sources = self.get_fpp_inputs(builder, context)
 
         if not sources:
-            print("[WARNING] No FPP inputs found in this module.")
+            print("[WARNING] No FPP sources found in this module.")
 
         # Build the input argument list
         input_args = []
@@ -146,7 +146,6 @@ class FppUtility(ExecutableAction):
 
         user_args = args[1]
         app_args = [self.utility] + user_args + input_args
-        # NOTE: This adds all fpp_inputs as source file arguments, not as --imports arguments
         if builder.cmake.verbose:
             print(f"[FPP] '{' '.join(app_args)}'")
         return subprocess.run(app_args, capture_output=False).returncode
--- a/src/fprime/fpp/common.py
+++ b/src/fprime/fpp/common.py
@@ -33,10 +33,14 @@ class FppUtility(ExecutableAction):
     in a similar order across utilities. This action executes these utilities via a subprocess using the command line
     format:
 
-    <utility> <user supplied args> <locations file> <fpp inputs>
+    <utility> <user supplied args> [<fpp imports>] <locations file> <fpp inputs>
+
+    Some fpp utilities distinguish between import files (--import flag) and source files. Those utilities should set
+    the imports_as_sources flag to False. This will cause the utility to pass the import files as --import arguments.
+    If imports_as_sources is True, the import files are passed as inputs just like source files.
     """
 
-    def __init__(self, name):
+    def __init__(self, name, imports_as_sources=True):
         """Construct this utility with the supplied name
 
         Args:
@@ -44,6 +48,7 @@ class FppUtility(ExecutableAction):
         """
         super().__init__(TargetScope.LOCAL)
         self.utility = name
+        self.imports_as_sources = imports_as_sources
 
     def is_supported(self, _=None, __=None):
         """Returns whether this utility is supported"""
@@ -67,7 +72,7 @@ class FppUtility(ExecutableAction):
         return locations_path
 
     @staticmethod
-    def get_fpp_inputs(builder: Build, context: Path) -> List[Path]:
+    def get_fpp_inputs(builder: Build, context: Path) -> Tuple[List[Path], List[Path]]:
         """Return the necessary inputs to an FPP run to forward to fpp utilities
 
         Returns two types of FPP files input into FPP utilities: the FPP files associated with the given module and the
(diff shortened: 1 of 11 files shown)


src/fprime/fpp/common.py
```diff
@@ -116,17 +129,24 @@ def execute(
 
         # Read files and arguments
         locations = self.get_locations_file(builder)
-        inputs = self.get_fpp_inputs(builder, context)
+        imports, sources = self.get_fpp_inputs(builder, context)
 
-        if not inputs:
+        if not sources:
             print("[WARNING] No FPP inputs found in this module.")
```
thomas-bc: Should say "No FPP sources" instead

src/fprime/fpp/impl.py
```diff
@@ -0,0 +1,154 @@
...
+
+    # Run fpp-to-cpp --template
+    FppUtility("fpp-to-cpp", imports_as_sources=False).execute(
+        build,
+        context,
+        args=(
+            {},
+            [
+                "--template",
+                *(["--unit-test"] if ut else []),
```
thomas-bc: Add option to pass in -a to fpp-to-cpp

src/fprime/util/cookiecutter_wrapper.py
```diff
@@ -36,43 +37,23 @@ def run_impl(build: Build, source_path: Path):
     hpp_dest = hpp_files[0]
     cpp_dest = common[0] if common else cpp_files[0]
```
thomas-bc: Remove common stuff. There is most likely a cleaner way to do all this since we know we'll only generate 2 files.

thomas-bc (commented):
Review comments

LeStarch:
There are two issues with this PR:

1. Running `fprime-util impl --ut` does not produce `.template` files, and thus will clobber the existing UT files
2. The imports aren't working for the SignalGen UT
  1. The import for `Ref/SignalGen/SignalGen.hpp` comes out as `SignalGen/SignalGen.hpp`
  2. The import in `SignalGenTestMain.cpp` isn't using a relative import, and thus cannot find `SignalGenTester.hpp`

These both might be issues in FPP (@bocchino, thoughts?)

thomas-bc:
(1) `fprime-util impl --ut` does not mess with filenames. Currently it seems FPP generates `<filename>.template` for `fpp-to-cpp`, but regular `<filename>` for `fpp-to-cpp --templates`. I can implement the file renaming, but we may want to implement that in FPP instead?
(2) and (3) are an oversight on my end - I forgot to test with the old project structure (e.g. Ref/)
(4) Not too sure there, may be related to (2) and (3), I'll investigate

LeStarch:
@thomas-bc:
(1) We should implement it in fpp
(4) This is likely an issue in FPP as I fixed a similar issue in fpp's Tester.cpp

bocchino:
> Running fprime-util impl --ut does not produce .template files

We can change this, but the way it works in FPP now is consistent with the way it works in XML and has always worked.

> and thus will clobber the existing UT files

I don't think so, because the existing UT files are typically in test/ut and the generated files are in the component directory. This way the user can just copy the files over, without renaming files. It is inconsistent with the .template naming for the flight implementation, but it seems to work well based on experience so far.

bocchino:
> This is likely an issue in FPP as I fixed a similar issue in fpp's Tester.cpp

I agree, I'll open an issue to fix it in FPP.

bocchino:
Is now a good time to make a v2.0.1 release of FPP to get these changes into F Prime mainline?

src/fprime/fpp/common.py
```diff
@@ -116,17 +129,23 @@ def execute(
...
+
+        # Build the input argument list
+        input_args = []
+        if self.imports_as_sources:
+            input_args = [
+                str(item) for item in itertools.chain([locations] + imports + sources)
+            ]
+        else:
+            input_args.extend(["-i", ",".join(map(str, imports))] if imports else [])
+            input_args.extend(sources)
```
LeStarch: Should this not be `[str(source) for source in sources]` to match the others?  Also, don't we need the `[locations]` bit?

src/fprime/fpp/impl.py
```diff
@@ -0,0 +1,163 @@
...
+        build.build_dir / "F-Prime",
+        build.build_dir,
+    ]
+
+    # Holds the list of generated files to be passed to clang-format
+    gen_files = tempfile.NamedTemporaryFile(prefix="fprime-impl-")
+
+    output_path = Path(output_dir)
+    if ut:
+        output_path = output_path / "test/ut"
```
LeStarch: This will cause clobbering. @bocchino recommends we generate in the component (not UT) directory until FPP adds `.template` to the UT impl outputs.  Then we can change.

LeStarch (changes requested):
Just two comments!

src/fprime/fpp/impl.py
thomas-bc: Sounds good to me. This will also keep it consistent with the current process, which I'm not opposed to. Meaning no rewriting of the tutorials etc...

src/fprime/fpp/common.py
thomas-bc: `subprocess.run()` does accept pathlib.Path objects, but googling it a little bit did show that this could lead to bugs every now and then (when debugging on specific versions of Windows? lol)... I'll change it for robustness's sake

src/fprime/fpp/impl.py
```diff
@@ -42,18 +42,16 @@ def fpp_generate_implementation(
     prefixes = [
         build.get_settings("framework_path", ""),
         *build.get_settings("library_locations", []),
-        build.cmake_root,
         build.build_dir / "F-Prime",
         build.build_dir,
     ]
+    if build.is_submodule_build_structure():
```
LeStarch: Why?    Also, this should not use "cmake root", but `build.get_settings("project_path")` should it not?

src/fprime/fpp/common.py
```diff
@@ -142,7 +142,8 @@ def execute(
             ]
         else:
             input_args.extend(["-i", ",".join(map(str, imports))] if imports else [])
-            input_args.extend(sources)
+            input_args.extend([str(source) for source in sources])
+            input_args.append(str(locations))
```
LeStarch: Does locations need to come first?  I always put it first, but perhaps it does not matter.

LeStarch (commented):
Left two more comments.

src/fprime/fpp/common.py
thomas-bc: Addressed

src/fprime/fpp/impl.py
thomas-bc: Yes absolutely, my mistake. By using project_root, we don't need to check whether the build structure is old or new, as project_root and framework_path will simply be the same. Note, both of these values are guaranteed to exist as per the following: 
https://github.com/fprime-community/fprime-tools/blob/4fc88dc0a20fcadd5b09bb5a188bc41373237d5d/src/fprime/fbuild/settings.py#L52-L53

since [find_fprime will error out](https://github.com/fprime-community/fprime-tools/blob/4fc88dc0a20fcadd5b09bb5a188bc41373237d5d/src/fprime/fbuild/settings.py#L36) if it can't find framework_location

LeStarch (approved):

LeStarch (approved):

PR #173, opened 2023-10-24
Fixing parallel builds

LeStarch:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes the parallel build `-jN` flag passing.  This has been broken since at least `v3.0.0`.  

## Rationale

Builds should be fast

- [x] Tested `fprime-util build -j16`
- [x] Tested `fprime-util check -j16`

Much faster.

## Future Work

None


LeStarch:
An explanation of the fix:

The job flag may be supplied to any target in make, however "all targets" were objects and the "command" was a string.  This checks against the mnemonic flags instead.

thomas-bc (approved):

PR #175, opened 2023-10-30
Remove Python 3.7 support

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**|  |

---
## Change Description

Sunsetting Python 3.7 support

## Rationale

Python 3.7 came to end of life: https://endoflife.date/python
Support has been dropped on https://github.com/nasa/fprime

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -72,15 +72,15 @@ to interact with the data coming from the FSW.
         "Operating System :: POSIX",
         "Programming Language :: Python",
         "Programming Language :: Python :: 3",
-        "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
         "Programming Language :: Python :: Implementation :: CPython",
         "Programming Language :: Python :: Implementation :: PyPy",
     ],
-    # Requires Python 3.7+
-    python_requires=">=3.7",
+    # Requires Python 3.8+
+    python_requires=">=3.8",
     install_requires=[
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -72,15 +72,15 @@
...
-        "Programming Language :: Python :: 3.7",
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
         "Programming Language :: Python :: 3.10",
+        "Programming Language :: Python :: 3.11",
         "Programming Language :: Python :: Implementation :: CPython",
         "Programming Language :: Python :: Implementation :: PyPy",
     ],
-    # Requires Python 3.7+
-    python_requires=">=3.7",
```
thomas-bc, line 83: Do we want this change?
This may cause confusion for users, as `pip install fprime-tools` will install older versions without warning if running Python 3.7. But I don't think there is a better solution. The alternative is to keep it in, which may cause errors when running pip install.

setup.py
```diff
@@ -72,15 +72,15 @@
...
+    # Requires Python 3.8+
+    python_requires=">=3.8",
```
thomas-bc, line 83: Do we want this change?
This may cause confusion for users, as `pip install fprime-tools` will install older versions without warning if running Python 3.7. But I don't think there is a better solution. The alternative is to keep it in, which may cause errors when running pip install.

LeStarch: When we break the bootstrapping script into its own entity that can support multiple versions of python and users of (deprecated) 3.7 could then install older versions of F´  with supported packages.

LeStarch (approved):

PR #176, opened 2023-11-02
Restricting visualization to current project if possible

LeStarch:
| | |
|:---|:---|
|**_Originating Project/Creator_**| Infra |
|**_Affected Component_**| fprime-util visualize |
|**_Related Issue(s)_**| nasa/fprime/#2354  |
|**_Has Unit Tests (y/n)_**| n/a |
|**_Builds Without Errors (y/n)_**| y  |
|**_Unit Tests Pass (y/n)_**| y |
|**_Documentation Included (y/n)_**| na/ |

---
## Change Description

Uses a projects inferred name to sub-select a topology to visualize when running with sub-topolgies!

## Rationale

Subtopologies make multiple `AppAi.xml` files but the user just wants to visualize the composite topology not the sub topologies.  Visualizing just the subtopology can be done in the subtopology's module.

### DIFF
--- a/src/fprime/fpp/visualize.py
+++ b/src/fprime/fpp/visualize.py
@@ -130,7 +130,7 @@ def run_fprime_visualize(
     config = {"SOURCE_DIRS": [str(viz_cache.resolve())]}
     app = construct_app(config)
     try:
-        webbrowser.open(f"http://localhost:{parsed.gui_port}")       
+        webbrowser.open(f"http://localhost:{parsed.gui_port}")
         app.run(port=parsed.gui_port)
     except KeyboardInterrupt:
         print("[INFO] CTRL-C received. Exiting.")
--- a/src/fprime/fpp/visualize.py
+++ b/src/fprime/fpp/visualize.py
@@ -6,6 +6,7 @@ import argparse
 import shutil
 import subprocess
 import tempfile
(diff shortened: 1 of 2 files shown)


LeStarch:
Also made it launch the webbrowser like the GDS does so that one doesn't need to know where to go look!

src/fprime/fpp/visualize.py
```diff
@@ -126,6 +130,7 @@ def run_fprime_visualize(
     config = {"SOURCE_DIRS": [str(viz_cache.resolve())]}
     app = construct_app(config)
     try:
+        webbrowser.open(f"http://localhost:{parsed.gui_port}")
```
thomas-bc, line 133: Should this come after app.run() ? Not that any browser would load that quickly... but just because?

LeStarch: No, because app.run will block until an interrupt.

LeStarch: To guarantee ordering, we'd need to spin-off a thread.  I am not sure that is worth it.

thomas-bc: Ohh right right, sounds good

thomas-bc (approved):

PR #177, opened 2023-11-07
Always print output when refreshing cache

### LINKED ISSUE #1992: F´ Util Prints "Error Not Found" Before Regenerating
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

When building a new component, `generate` is triggered via `noop` build.  However, the "not found" error prints once, before this happens confusing users.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/1992 |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/1992

When `cmake_refresh_cache()` is triggered, always print the cmake regeneration output.

## Rationale

- This is consistent with the behavior that `fprime-util build` et al. have when target **exists** but requires a refresh (e.g. `.fpp` file was modified). Output of regeneration is printed out.
- Imo it helps new users (who may not be familiar with CMake) in understanding the build process. Target doesn't exist -> needs a refresh. Target is out of date -> needs a refresh. Refresh is the same thing happening under the hood in both cases (also `fprime-util generate`), and the familiar-looking output confirms that.
- Clears out any confusion with regards to https://github.com/nasa/fprime/issues/1992

## Comment
This has the side effect of also printing the `Built target noop` to the screen... which I don't really like.
I've been looking in other ways to do this, but was unsuccessful. A fix could be to go in the [CMakeHandler._communicate](https://github.com/thomas-bc/fprime-tools/blob/devel/src/fprime/fbuild/cmake.py#L566-L568) and explicitly skip lines that equal `Built target noop\n`, but that sounds rather ugly... 
Do you have any suggestion?

### DIFF
--- a/src/fprime/fbuild/cmake.py
+++ b/src/fprime/fbuild/cmake.py
@@ -306,6 +306,8 @@ class CMakeHandler:
         Returns:
             path string that is relative to some root of the project. i.e. used for target suffix names
         """
+        if path is None:
+            path = os.getcwd()
         # Get module name from the relative path to include root
         include_root = self.get_include_info(path, build_dir)[1]
         return os.path.relpath(path, include_root)
@@ -340,6 +342,17 @@ class CMakeHandler:
             for make in self.cached_help_targets
             if make.startswith(prefix)
         ]
+
+    def is_target_supported(self, build_dir: str, target: str):
+        """Checks if a target is supported by the current build directory
+
+        Args:
+            build_dir: build directory to use for detecting targets
+            target: target name to check
+        """
+        # Load in self.cached_help_targets
+        self.get_available_targets(build_dir, None)
+        return target in self.cached_help_targets
 
     @staticmethod
     def purge(build_dir):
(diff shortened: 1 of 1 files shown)


LeStarch (approved):

LeStarch:
We could rename the target to "refresh".

LeStarch (changes requested):
Requires the following:
- Run "refresh_cache" instead of "noop" if it exists, else run "noop"

src/fprime/fbuild/cmake.py
```diff
@@ -439,19 +452,30 @@ def cmake_refresh_cache(self, build_dir, full=False):
...
-            self.execute_known_target(
-                "noop",
-                build_dir,
-                None,
-                top_target=True,
-                full_cache_rebuild=True,
-                print_output=self.verbose,
-            )
+            # Backwards compatibility: refresh_cache was named noop until v3.3.x
+            if self.is_target_supported(build_dir, "noop"):
```
LeStarch, line 461: Given the slowness of printing targets, perhaps it is faster to run "refresh_cache" and trap "No rule to make target `llkjldksjlksdj'.  Stop." and then run "noop"?

LeStarch: Will revisit with the ninja support work.

LeStarch (approved):

PR #181, opened 2023-12-06
Add pyproject.toml

thomas-bc:
| | |
|:---|:---|
|**_Affected Component_**|  |
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/1995 |

---
## Change Description

- Add `pyproject.toml` 
- Strip setup.py from old config
- Update publishing workflow to use trusted publishers mechanism

## Rationale

Conform to PEP517, 518 et al. Modernize our Python setup so as not to be left behind. Potentially help with Python 3.12 support

Tested in branch with test.pypi.org: 
https://github.com/fprime-community/fprime-tools/actions/runs/7119122735
https://test.pypi.org/project/fprime-tools/0.0.1a2/

TODO:
- sanity check that this switch won't break other external things?
- alpha release and let CI fly


.github/workflows/publish.yml
```diff
@@ -1,25 +1,29 @@
...
-        TWINE_PASSWORD: ${{ secrets.PYPI_CREDENTIAL }}
-      with:
-        repo: "pypi"
-        package: "fprime-tools"
+      - name: "Checkout Repository"
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Setup Python
+        run: pip3 install -U pip wheel build setuptools setuptools_scm twine
```
LeStarch: Don't we just need `build` and `twine` as it should detect the other necessary tools from the requires section.

LeStarch (changes requested):
One comment in-line.  Additionally `setup.py` should now look like:

```
from setuptools import setup

setup()
```
As noted for backwards compatibility here: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html

.github/workflows/publish.yml
thomas-bc: Good point

LeStarch (dismissed):
PY 12?  but not a blocker.

LeStarch (approved):

PR #188, opened 2024-01-05
Add Ninja support

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2376 |
|**_Has Unit Tests (y/n)_**| CI int test |

---
## Change Description

Add support for using the Ninja CMake Generator. Turns out not a lot of things needed changing, for one main reason: Ninja refreshes the build cache automatically when it detects changes or a missing target. 

`fprime-util generate --ninja` generates a build cache using Ninja.

Beta release? I tested manually as much as I could, but I'm sure there's going to be edge cases.

### DIFF
--- a/src/fprime/fbuild/cli.py
+++ b/src/fprime/fbuild/cli.py
@@ -214,7 +214,7 @@ def add_special_targets(
     generate_parser.add_argument(
         "--ninja",
         action="store_true",
-        help="Uses Ninja as the build system.",
+        help="Use the Ninja build system instead of Unix Makefiles",
     )
     # The following option is specified only to show up in --help.
     # It is not handled by argparse, but in fprime.util.cli:validate()
--- a/src/fprime/fbuild/cli.py
+++ b/src/fprime/fbuild/cli.py
@@ -59,6 +59,8 @@ def run_fbuild_cli(
             cmake_args["ENABLE_SANITIZER_UNDEFINED_BEHAVIOR"] = "ON"
         if toolchain is not None:
             cmake_args["CMAKE_TOOLCHAIN_FILE"] = toolchain
+        if parsed.ninja:
+            cmake_args["CMAKE_GENERATOR"] = "Ninja"
         build.generate(cmake_args)
     elif parsed.command == "purge":
         # Since purge does not load its "base", we need to overload the platform
@@ -208,6 +210,11 @@ def add_special_targets(
         default=False,
         help="Disable the compiler sanitizers. Sanitizers are only enabled by default when --ut is provided.",
         action="store_true",
+    )
+    generate_parser.add_argument(
+        "--ninja",
+        action="store_true",
+        help="Uses Ninja as the build system.",
     )
     # The following option is specified only to show up in --help.
     # It is not handled by argparse, but in fprime.util.cli:validate()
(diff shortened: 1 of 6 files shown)


src/fprime/fbuild/cmake.py
```diff
@@ -97,6 +97,13 @@ def execute_known_target(
         self.validate_cmake_cache(cmake_args, build_dir)
 
         make_args = {} if make_args is None else make_args
+
+        run_args = ["--build", build_dir]
+        # CMake 3.12+ supports parallel builds with -j
+        jobs = make_args.pop("--jobs", None)
+        if jobs is not None:
+            run_args.extend(["-j", str(jobs)])
```
LeStarch, line 105: Note: try to not specify the -j flag when building without it.

src/fprime/fbuild/cmake.py
```diff
@@ -330,11 +341,25 @@ def get_available_targets(self, build_dir, path):
             stdout, _ = self._run_cmake(
                 run_args, write_override=True, print_output=False
             )
-            self.cached_help_targets = [
-                line.replace("...", "").strip()
-                for line in stdout
-                if line.startswith("...")
-            ]
+            # help target lists all targets but has a different output format for ninja and make
```
LeStarch, line 344: Can this function be removed entirely?

LeStarch: I think this was determined as necessary.

src/fprime/util/cli.py
```diff
@@ -242,7 +242,8 @@ def validate(parsed, unknown):
     # Build type only for generate, jobs only for non-generate
     elif parsed.command in [target.mnemonic for target in Target.get_all_targets()]:
         parsed.settings = None  # Force to load from cache if possible
-        make_args["--jobs"] = 1 if parsed.jobs <= 0 else parsed.jobs
+        if parsed.jobs:
```
LeStarch: `parsed.jobs is None` otherwise `0` will skip this check and miss the `make_args["--jobs"] = 1 if parsed.jobs <= 0 else parsed.jobs` correction

src/fprime/fbuild/cli.py
```diff
@@ -143,9 +143,8 @@ def add_target_parser(
         parser.add_argument(
             "-j",
             "--jobs",
-            default=1,
```
LeStarch, line 144: default=None?  or is this implicit?

src/fprime/util/cli.py
LeStarch: Now that I think about it, we should have

```
if parsed.jobs is None or parsed.jobs <=0:
    make_args["--jobs"] = parsed.jobs
....

LeStarch: That way --jobs is only set when valid, otherwise default.

LeStarch (changes requested):

src/fprime/fbuild/cli.py
thomas-bc: It's implicit - `default`'s default value is None

src/fprime/util/cli.py
thomas-bc: Updated

LeStarch (approved):

LeStarch (approved):

PR #192, opened 2024-01-23
Drop AcConstants.ini support

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Removes handling of the variables related to the old AcConstants.ini file

## Rationale

See https://github.com/nasa/fprime/pull/2496


thomas-bc:
Actually - this likely hurts backwards compatibility, while not being hurtful if we were to keep it in... 
So maybe we should just keep it in?

Thoughts @LeStarch ?

LeStarch (approved):

PR #197, opened 2024-03-09
CR 2375: TimeType.to_readable should display microsecond precision.

Ahmad-Bamba:
| | |
|:---|:---|
|**_Related Issue(s)_**| [2375](https://github.com/nasa/fprime/issues/2375) |
|**_Documentation Included (y/n)_**| n  |

---
## Change Description

This change makes the TimeType.to_readable function use datetime.isoformat to convert the datetime object to a human readable format. In doing so, fprime-gds outputs `channel.log` entries with microsecond precision.

## Rationale

> The fundamental philosophy that must be applied to the ground system is that any conversion (time, DN->EU, units, etc.) should never decrease the resolution of the source data.

I tested the changes locally by first following the [fprime hello world](https://fprime-community.github.io/fprime-tutorial-hello-world/) tutorial. Then I installed the local version of fprime-tools, and ran the example project with `fprime-gds`. This produced the following output that I've attached below showing no more accuracy loss in `channel.log`:

image

Compared with before this change:

image

## Future Work

This seems like too a small change to make a unit test for, but I could think of a good way to design one if requested.

### DIFF
--- a/.github/actions/spelling/expect.txt
+++ b/.github/actions/spelling/expect.txt
@@ -115,6 +115,7 @@ isdir
 isfile
 isinstance
 isnumeric
+isoformat
 Isr
 issubclass
 iterdir
@@ -246,6 +247,7 @@ testbuild
 testimpl
 textwrap
 timebase
+timespec
 toctree
 todo
 toolchain
--- a/src/fprime/common/models/serialize/time_type.py
+++ b/src/fprime/common/models/serialize/time_type.py
@@ -289,7 +289,10 @@ class TimeType(type_base.BaseType):
 
         # If we could convert to a valid datetime, use that, otherwise, format
         if dt:
-            return dt.strftime("%Y-%m-%d %H:%M:%S%z")
+            # datetime.isoformat() returns time string with microsecond 
+            # precision.
+            # This line can be changed for other precisions or needs.
+            return dt.isoformat()
         return "%s: %d.%06ds, context=%d" % (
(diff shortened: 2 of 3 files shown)


thomas-bc (changes requested):
Looks good to me! Thank you!

Could we just force the timespec `dt.isoformat(timespec="microseconds")` ? Default seems to be `auto` which falls back to microseconds in our case, but you never know, and it's also easier to read/understand that way.

You're also going to need to [format the file](https://github.com/fprime-community/fprime-tools/tree/devel?tab=readme-ov-file#black-formatter) for it to pass CI, and add `isoformat` to https://github.com/fprime-community/fprime-tools/blob/devel/.github/actions/spelling/expect.txt for spelling

Ahmad-Bamba:
Feel free to let me know if there is anything else that can be done to improve this

thomas-bc:
@Ahmad-Bamba could you please also push the timespec="microseconds" change?

Ahmad-Bamba:
Ah, I neglected to actually git add that file. My bad

Ahmad-Bamba:
Should be good. Sorry about that!

thomas-bc (approved):

thomas-bc:
@LeStarch would you put a second set of eyes on this ?

LeStarch (approved):
This is a *great* solution.  Much better than what we had.  I confirmed that from Python 3.6 through 3.11 the documentation backs-up the change (this call is supported in all versions)!

PR #207, opened 2024-06-28
Add fpp-to-json utilities

mosa11aei:
| | |
|:---|:---|
|**_Related Issue(s)_**| n |

---
## Change Description

The following PR implements some utilities for using `fpp-to-json` with Python, specifically an abstract syntax tree (AST) converter and an FPP writer. The utilities have been made broad enough to be able to add visitors for different AST output types (i.e., XML).


mosa11aei:
Do not merge yet; still needs a bit of refactoring.

LeStarch (dismissed):
LGTM

thomas-bc (commented):
Would it be useful to add an entrypoint to fprime-util to have e.g. `fprime-util fpp-to-json` ? Or is the goal to generate the JSON stuff in the build cache as part of the build process?

mosa11aei:
> Would it be useful to add an entrypoint to fprime-util to have e.g. `fprime-util fpp-to-json` ? Or is the goal to generate the JSON stuff in the build cache as part of the build process?

So this utility is a parser and data type converter from the abstract syntax tree to Python. Because of this, it is not a CLI that you run to use fpp-to-json, rather you import the visitors and apply them to a given AST (which can be obtained any which way one wants).

In general it may be nice to include a fpp-to-json entry point, which calculates dependencies alongside running the analysis. What use case are you thinking?

bocchino:
> Would it be useful to add an entrypoint to fprime-util to have e.g. fprime-util fpp-to-json ? 

I don't think fpp-to-json is a user-facing tool. It's a tool that developers can use to build other tools.

thomas-bc (approved):

PR #208, opened 2024-07-02
Swap `regCommands()` and `configTopology` + Move Buffmgr and Framing setup to front

Lex-ari:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime#2791 |

---
## Change Description

Swap `regCommands()` and `configureTopology()` in Topology.cpp
Move BufferManager Setup and Framing/Deframing setup to front in `configureTopology()` in Topology.cpp

## Rationale
* Registering commands also generates events as a byproduct. In a hub pattern implementation connected to eventLogging, the hub will assert since it tries to handle the events before `configureTopology()` is called (and the bufferManager is not setup).
* There may be other similar implementations from other projects that require a more specific setup before their components are ready to handle commands and events.
* Have not found a use case to handle commands before the system setup is completed.

* BufferManager and Framing/Deframing should be setup before `readParamFile()` is called. This also sends an event and the hub would assert before the bufferManager and deframer is setup.

CI to ensure standard projects are working correctly

CC: @timcanham @LeStarch @bocchino @Joshua-Anderson

### DIFF
--- a/src/fprime/cookiecutter_templates/cookiecutter-fprime-deployment/{{cookiecutter.deployment_name}}/Top/{{cookiecutter.deployment_name}}Topology.cpp
+++ b/src/fprime/cookiecutter_templates/cookiecutter-fprime-deployment/{{cookiecutter.deployment_name}}/Top/{{cookiecutter.deployment_name}}Topology.cpp
@@ -84,6 +84,21 @@ Svc::Health::PingEntry pingEntries[] = {
  * desired, but is extracted here for clarity.
  */
 void configureTopology() {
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):
The code change itself LGTM. I'll let @LeStarch confirm this is something we want to change

timcanham:
LGTM to me too. Order shouldn't matter.

LeStarch (approved):
I approved this contingent upon @timcanham's approval, which we have.

LeStarch:
@Lex-ari: we need to also change the bootstrap templates.

PR #206, opened 2024-06-11
Uncomment noop functions from topology template

mosa11aei:
Uncomment parameters call (ac) and include config components for phases (both noops out of the box)

| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2774  |
|**_Has Unit Tests (y/n)_**| n |
|**_Documentation Included (y/n)_**| n |

---
## Change Description

Certain functions (i.e. `loadParameters`) are now noops within F Prime. However, they are commented out of the box in the topology template files, which is inconvenient. They are now uncommented.

PR #212, opened 2024-07-29
Adding phase configure to manual cookiecutter. nasa/fprime#2820

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fixes #2820.

### DIFF
--- a/src/fprime/cookiecutter_templates/cookiecutter-fprime-deployment/{{cookiecutter.deployment_name}}/Top/{{cookiecutter.deployment_name}}Topology.cpp
+++ b/src/fprime/cookiecutter_templates/cookiecutter-fprime-deployment/{{cookiecutter.deployment_name}}/Top/{{cookiecutter.deployment_name}}Topology.cpp
@@ -149,8 +149,6 @@ void setupTopology(const TopologyState& state) {
     configureTopology();
     // Autocoded command registration. Function provided by autocoder.
     regCommands();
-    // Component-specific configurations using fpp phases. Function provided by autocoder.
-    configComponents(state);
     // Autocoded parameter loading. Function provided by autocoder.
     loadParameters();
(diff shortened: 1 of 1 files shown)


src/fprime/cookiecutter_templates/cookiecutter-fprime-deployment/{{cookiecutter.deployment_name}}/Top/{{cookiecutter.deployment_name}}Topology.cpp
```diff
@@ -143,6 +143,8 @@ void setupTopology(const TopologyState& state) {
     setBaseIds();
     // Autocoded connection wiring. Function provided by autocoder.
     connectComponents();
+    // Autocoded configuration. Function provided by autocoder.
+    configComponents(state);
```
thomas-bc, line 147: @LeStarch this has been fixed in https://github.com/nasa/fprime-tools/pull/206/files already

Should the call be made before `configureTopology()` ? In which case it should be removed line 153 (further below)

LeStarch: Yes, I believe so.  The phase code cannot, by definition, know or care what the hand config does.  As such, I assumed it was safest to place this first.

LeStarch: @thomas-bc fixed!  What was the reason to place this after register commands?

thomas-bc (approved):

PR #216, opened 2024-09-12
Remove restrictions for the Build class to setup its platform

### LINKED ISSUE #2869: fprime-util generate should not fail if run twice
| | |
|:---|:---|
|**_F´ Version_**| 3.4.3|
|**_Affected Component_**| N/A  |
---
## Problem Description

fprime-util generate fails if it has already been run, even if the first run was successful. 

A typical cmake workflow is ```cmake .. && make```. The parallel for F' would be ```fprime-util generate && fprime-util build```, which fails after the first generate. Understood that fprime-util build probably re-runs generate if it needs to, and you only _need_ to run generate once, but from a 'use tools like you're used to' perspective, fprime-util generate should be indempotent and not fail if it has nothing to do.

## Context / Environment

Execute `fprime-util version-check` and share the output.

```
$ fprime-util version-check
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-6.8.0-40-generic-x86_64-with-glibc2.35
Python version: 3.10.12
CMake version: 3.22.1
Pip version: 24.2
Pip packages:
    fprime-tools==3.4.4
    fprime-gds==3.4.3
    fprime-fpp-*==2.1.0a3

```

## How to Reproduce

Run fprime-util generate twice

## Expected Behavior

Similar to cmake, the second time should either print 'already generated', or some other small message and return success. This is useful for scripting 'generate' and 'build' tasks which rely on the error code.

weggert2:
Which allows users to call fprime-util generate several times without faiure

| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2869 |
|**_Has Unit Tests (y/n)_**|  yes |
|**_Documentation Included (y/n)_**| yes |

---
## Change Description

Removes restrictions on calling Build.invent and Build.__setup_default more than once. I admit I don't understand the full implications of this, so I'd encourage the author @LeStarch to evaluate whether this breaks things. Anecdotally, locally installing this allows me to call generate in my fprime project several times without error.

## Rationale

Fixes a (perceived) bug that fprime-util generate should not fail if a cache already exists

There is a very simple included unit test that shows that calling Builder.invent() twice does not raise an assertion. Please let me know what else I need to test.

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -72,8 +72,8 @@ class Build:
     def invent(self, platform: str = None, build_dir: Path = None):
         """Invents a build path from a given platform
 
-        Sets this build up as a new build that would be used as as part of a generate step. This directory must not
-        already exist. If platform is None, a default will be chosen from the settings.ini file. If the settings.ini
+        Sets this build up as a new build that would be used as as part of a generate step.
+        If platform is None, a default will be chosen from the settings.ini file. If the settings.ini
         file does not exist, or does not specify a default_toolchain, then "native" will be used. Settings are loaded in
         this step for further uses of this build.
 
@@ -84,14 +84,8 @@ class Build:
             platform:   name of platform to build against. None will use default from settings.ini or without this
                         setting, "native". Defaults to None.
             build_dir:  explicitly sets the build path to allow for user override of default
-
-        Raises:
-            InvalidBuildCacheException: a build cache already exists as it should not
         """
         self.__setup_default(platform, build_dir)
-        if self.build_dir.exists():
-            msg = f"{self.build_dir} already exists."
-            raise InvalidBuildCacheException(msg)
 
     def load(self, platform: str = None, build_dir: Path = None, skip_validation=False):
         """Load an existing build cache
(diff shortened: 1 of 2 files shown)


thomas-bc:
It looks like this re-generates regardless of whether there is a need to. Isn't CMake's default behavior to skip the regeneration?

From quickly looking at https://github.com/weggert2/fprime-tools/blob/5d964ecefcbc4010d8bb0b5d3290687510233530/src/fprime/fbuild/builder.py#L365-L366 and the cmake.generate_build() , I am not too sure why this happens since we delegate to CMake

thomas-bc (commented):
It looks like this re-generates regardless of whether there is a need to. Isn't CMake's default behavior to skip the regeneration?

From quickly looking at https://github.com/weggert2/fprime-tools/blob/5d964ecefcbc4010d8bb0b5d3290687510233530/src/fprime/fbuild/builder.py#L365-L366 and the cmake.generate_build() , I am not too sure why this happens since we delegate to CMake

thomas-bc:
@weggert2 we went ahead with the implementation of `fprime-util generate --force` flag. This will purge and re-generate in a single command. The reason for not doing a re-generation without purging is that autocoded files can be leftover in the build cache, or not tracked correctly, which could cause issues.

Thank you for the contribution nonetheless! It brought up a good use-case, we just went with that alternate approach.

weggert2:
Sounds good to me! Thanks for tackling that!

________________________________

PR #217, opened 2024-09-20
add force option to generate that will call purge before generating, …

csmith608:
…still need to check edgecases

| | |
|:---|:---|
|**_Related Issue(s)_**| https://github.com/nasa/fprime/issues/2887 |

---
## Change Description

Implements force generate the runs a purge

## Rationale

QoL

fprime-util generate --force

### DIFF
--- a/src/fprime/fbuild/cli.py
+++ b/src/fprime/fbuild/cli.py
@@ -51,7 +51,8 @@ def run_fbuild_cli(
     """
     if parsed.command == "generate":
         if parsed.force:
-            purge(build, parsed)
+            print(f"[INFO] Purging build directory at: {build.build_dir}")
+            build.purge()
 
         toolchain = build.find_toolchain()
         print(f"[INFO] Generating build directory at: {build.build_dir}")
--- a/src/fprime/fbuild/cli.py
+++ b/src/fprime/fbuild/cli.py
@@ -50,7 +50,7 @@ def run_fbuild_cli(
         make_args: arguments to supply to the build tool (make or ninja)
     """
     if parsed.command == "generate":
-        if(parsed.force): 
+        if parsed.force:
             purge(build, parsed)
 
         toolchain = build.find_toolchain()
@@ -69,7 +69,6 @@ def run_fbuild_cli(
     elif parsed.command == "purge":
         # Since purge does not load its "base", we need to overload the platform
         purge(build, parsed)
-        
 
     else:
         target = get_target(parsed)
@@ -83,10 +82,8 @@ def run_fbuild_cli(
             args=(make_args, pass_through, option_args),
         )
 
-def purge(
-    build: Build,
-    parsed: argparse.Namespace
-):
+
+def purge(build: Build, parsed: argparse.Namespace):
     for purge_build in Build.get_build_list(
         build, parsed.build_cache, ignore_invalid=parsed.force
     ):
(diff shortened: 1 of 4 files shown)


csmith608:
@LeStarch

csmith608:
@thomas-bc

LeStarch (commented):
What happens when you purge -f  then generate -f.  That will purge a second time, does that error?

Otherwise this looks great!

csmith608:
@LeStarch that's fine because the generate -f goes through all the purge logic so it realized those directories don't exist. 

This is the result from generate -f 

```
[INFO] Purge build directory at: /Users/crsmith/Documents/fprime/Ref/build-fprime-automatic-native
[INFO] Purge build directory at: /Users/crsmith/Documents/fprime/Ref/build-fprime-automatic-native-ut
[INFO] Skipping purge. The following directory does not exist: /Users/crsmith/Documents/fprime/Ref/build-fprime-automatic-native-ut
```

LeStarch (dismissed):

src/fprime/fbuild/cli.py
```diff
@@ -118,6 +83,47 @@ def run_fbuild_cli(
         )
 
 
+def purge(build: Build, parsed: argparse.Namespace):
+    for purge_build in Build.get_build_list(
+        build, parsed.build_cache, ignore_invalid=parsed.force
+    ):
+        print(
+            f"[INFO] {parsed.command.title()} build directory at: {purge_build.build_dir}"
```
thomas-bc, line 92: This causes some unexpected output
```
[INFO] Generate build directory at: fprime/Ref/build-fprime-automatic-native
[INFO] Generate build directory at: fprime/Ref/build-fprime-automatic-native-ut
[INFO] Skipping purge. The following directory does not exist: fprime/Ref/build-fprime-automatic-native-ut
[INFO] Generating build directory at: fprime/Ref/build-fprime-automatic-native
[INFO] Using toolchain file None for platform default
-- The C compiler identification is AppleClang 16.0.0.16000026
```

Since we're purging with the `parsed.command` = `generate`
The easiest is probably to hardcode the string, e.g. `"[INFO] Purging build dir...."

src/fprime/fbuild/cli.py
```diff
@@ -50,6 +50,9 @@ def run_fbuild_cli(
         make_args: arguments to supply to the build tool (make or ninja)
     """
     if parsed.command == "generate":
+        if parsed.force:
+            purge(build, parsed)
```
thomas-bc: The `build` object has a `purge()` method that basically just calls `rm -rf` on that specific build. I wonder if it would be better to use that (maybe with an additional info log). Otherwise `fprime-util generate --force` will wipe all build caches, and not just the one we are trying to force-generate.

src/fprime/fbuild/cli.py
thomas-bc: (This would not need fixing if we were to go with the below suggestion)

thomas-bc (changes requested):
One suggestion

src/fprime/fbuild/cli.py
LeStarch: Agreed.  Can we use that?

LeStarch (approved):

thomas-bc (approved):

PR #227, opened 2024-11-22
Fixing sanitizer handling. Fixes nasa/fprime#2678

### LINKED ISSUE #2678: Sanitizer Disable Does Not Work
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Generating with `-DENABLE_SANITIZER_ADDRESS=OFF` does not work.  Only the `--disable-sanitizer` to disable everything.  The culprit is:

https://github.com/fprime-community/fprime-tools/blob/c15253c98ffc816bcca24db43f24c4c070b7c5ff/src/fprime/fbuild/cli.py#L55-L59

> Notice how the sanitizers are forced-on. 

The fix is in several stages:
1. In CMake default to sanitizers should be == `BUILD_TESTING` (default "ON" when testing).
2. Above code should read "if disable" set the ENABLES "OFF"
3. Take no action on disable sanitizers not set

This way the sanitizers are explicitly turned off, or left to defaults.

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This fixes nasa/fprime#2678.  Essentially, the presence of the disable sanitizer flag should disable the sanitizers. In other cases, the sanitizer handling should be left to the underlying fprime/CMake implementation.

### DIFF
--- a/src/fprime/fbuild/cli.py
+++ b/src/fprime/fbuild/cli.py
@@ -57,11 +57,11 @@ def run_fbuild_cli(
         toolchain = build.find_toolchain()
         print(f"[INFO] Generating build directory at: {build.build_dir}")
         print(f"[INFO] Using toolchain file {toolchain} for platform {parsed.platform}")
-        if parsed.ut and not parsed.disable_sanitizers:
+        if parsed.disable_sanitizers:
             # The following options are defined in F' to have CMake enable the sanitizers
-            cmake_args["ENABLE_SANITIZER_LEAK"] = "ON"
-            cmake_args["ENABLE_SANITIZER_ADDRESS"] = "ON"
-            cmake_args["ENABLE_SANITIZER_UNDEFINED_BEHAVIOR"] = "ON"
+            cmake_args["ENABLE_SANITIZER_LEAK"] = "OFF"
+            cmake_args["ENABLE_SANITIZER_ADDRESS"] = "OFF"
+            cmake_args["ENABLE_SANITIZER_UNDEFINED_BEHAVIOR"] = "OFF"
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):

thomas-bc (changes requested):
.

src/fprime/fbuild/cli.py
```diff
@@ -57,11 +57,11 @@ def run_fbuild_cli(
         toolchain = build.find_toolchain()
         print(f"[INFO] Generating build directory at: {build.build_dir}")
         print(f"[INFO] Using toolchain file {toolchain} for platform {parsed.platform}")
-        if parsed.ut and not parsed.disable_sanitizers:
+        if parsed.disable_sanitizers:
```
thomas-bc, line 60: This line is the only place in the system (currently) where sanitizers do get enabled. Without it, sanitizers will always be disabled unless the `ENABLE_SANITIZERS_*` flags are passed to CMake.

Is this the intended behavior? It feels to me that sanitizers on UTs are a good practice.

Not sure why nasa/fprime#2678 fails though...

thomas-bc: Oh I should have read nasa/fprime#2678 prior to posting this lol

Are you intending to make a PR to F´ to have `SANITIZERS_* = BUILD_TESTING`

LeStarch: My intent was to set sanitizers in CMake in F´, yes.

thomas-bc (approved):

PR #235, opened 2024-12-18
Pass Environment To CMake in all Cases

### LINKED ISSUE #2033: Build F´ Targets Wrapper
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

Make a convience helper in fprime util for building targets directly.  E.g. Replace `cd <build>; make <target>` with `fprime-util build --target <target>`.

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

When `settings.ini` defines an environment, it needs to ensure that that environment is passed to CMake in all cases (with the exception of the `cmake --help` test.

## Rationale

Fixes nasa/fprime#2033

Added a number of UTs to test key functionality in `cmake.py`

## Future Work

Environment should be passed to wrapped utilities too (fpp, etc).  This will require more work: https://github.com/nasa/fprime/issues/3081

This extra work is not the cause of the reported issue.

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -440,6 +440,18 @@ class Build:
         path = Path(self.settings["install_destination"])
         return path if path.exists() else None
 
+    def refresh(self):
+        """ Refresh this build (i.e. the cmake build cache)
+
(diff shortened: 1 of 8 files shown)


LeStarch:
Closing, incorrect branch and issue.

PR #239, opened 2025-01-30
Update `fprime-util visualize` to use `fpp-to-layout`

jwest115:
| | |
|:---|:---|
|**_Related Issue(s)_**|   https://github.com/nasa/fprime/issues/3142 |

---
## Change Description

Updated the `fprime-util visualize` utility to use the `fpp-to-layout` tool instead of topology XML files.

## Rationale

Addresses https://github.com/nasa/fprime/issues/3142.

Tested against Ref deployment and added a second topology to test the multiple topology case. Confirmed the expected JSON files were generated and topology visualization looked as expected.

Ref Uplink connection graph:
image

Ref Topology (all connection graphs in the topology):
image

Topology 2 (topology with a single connection graph)
image

### DIFF
--- a/src/fprime/fpp/visualize.py
+++ b/src/fprime/fpp/visualize.py
@@ -74,33 +74,39 @@ def run_fprime_visualize(
             ["--directory", str(txt_cache)],
         ),
     )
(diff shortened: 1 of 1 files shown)


thomas-bc (approved):
Thank you! 🚀

PR #243, opened 2025-02-25
Removing Cheetah3

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

A description of the changes contained in the PR.

## Rationale

A rationale for this change. e.g. fixes bug, or most projects need XYZ feature.

Fill in testing procedures, specific items to focus on for review, or other info to help the team verify these changes are flight-quality.

## Future Work

Note any additional work that will be done relating to this issue.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -13,9 +13,13 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
         fprime-version: ["v3.4.3", "v3.5.1", "v3.6.0", "devel"]
-
+        exclude:
+        - python-version: 3.13
+          fprime-version: v3.4.3
+        - python-version: 3.13
+          fprime-version: v3.5.1          
     steps:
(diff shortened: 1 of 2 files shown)


thomas-bc (dismissed):

.github/workflows/integration-tests.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]
+        python-version: ["3.8", "3.9", "3.10", "3.11", "3.12"]
```
thomas-bc: Can we use [`exclude`](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#excluding-matrix-configurations) to exclude 3.13 runs from combining with older versions of F´ that break on it - rather than not running 3.13 at all ?

thomas-bc (changes requested):

.github/workflows/integration-tests.yml
thomas-bc:
```
    exclude:
      - python-version: 3.13
        fprime-version: 3.8
[...]
```

thomas-bc (approved):

PR #247, opened 2025-03-25
Fixes fprime-util purge error on broken directory

### LINKED ISSUE #3033: fprime-util purge does not always work
| | |
|:---|:---|
|**_F´ Version_**| v3.5.0-22-g00c5c8e11 |
|**_Affected Component_**| build system |
---
## Problem Description

I interrupted the generate command with Ctrl-C, and now I have a build directory that the purge command won't remove.

```
# fprime-util purge
[WARNING] Build cache '[...]/build-fprime-automatic-[...]' invalid or not found. Skipping.
[WARNING] Build cache '[...]/build-fprime-automatic-[...]-ut' invalid or not found. Skipping.
# fprime-util generate --ninja
[ERROR] [...]/build-fprime-automatic-[...] already exists.
```

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-5.15.167.4-microsoft-standard-WSL2-x86_64-with-glibc2.34
Python version: 3.9.18
CMake version: 3.26.5
Pip version: 21.2.3
Pip packages:
    fprime-tools==3.5.0
    fprime-gds==3.5.0
    fprime-fpp-*==2.2.1a1
```

## How to Reproduce

1. Run `fprime-util generate --ninja` and hit Ctrl-C partway through.
2. Run `fprime-util purge`
3. Run `fprime-util generate --ninja`

## Expected Behavior

`fprime-util purge` should remove the build directory even if it is broken, so that `fprime-util generate --ninja` can generate a clean build directory.

tylerrussin:
| | |
|:---|:---|
|**_Related Issue(s)_**| N/A |

---
## Change Description

The bug occurs when the fprime-util generate command is interrupted (such as with Ctrl+C) before the directory can finish building. This results in the fprime-util purge command to error out since it is unable to verify if the directory it is trying to delete is an F Prime directory.

The proposed fix adds an empty file early on in the directory's build process, allowing the purge command to identify that the directory it is trying to delete is an F Prime directory. New logic has also been added to allow the purge command to delete directories that are empty.  

## Rationale

This change Fixes nasa/fprime#3033

### DIFF
--- a/src/fprime/fbuild/builder.py
+++ b/src/fprime/fbuild/builder.py
@@ -94,6 +94,7 @@ class Build:
             if not force:
                 raise InvalidBuildCacheException(msg)
 
+  
     def load(self, platform: str = None, build_dir: Path = None, skip_validation=False):
         """Load an existing build cache
 
@@ -110,31 +111,29 @@ class Build:
             InvalidBuildCacheException: the build cache does not exist as it must
         """
         self.__setup_default(platform, build_dir)
-        # We only load empty directories or directories that contain the .fprime-build-dir file
-        if not skip_validation and (
-            not self.build_dir.exists()
-            or (
-                len(os.listdir(self.build_dir)) > 0
-                and not (self.build_dir / ".fprime-build-dir").exists()
+        
+        if skip_validation:
+            return
+        if self.build_dir.exists() and len(os.listdir(self.build_dir)) == 0 or (self.build_dir / ".fprime-build-dir").exists():
+            return
+        
+        # Message for hard-supplied --build-cache message
+        if build_dir is not None:
+            gen_args = f" --build-cache {build_dir}"
+        else:
+            gen_args = " --ut" if self.build_type == BuildType.BUILD_TESTING else ""
+            gen_args += (
+                " " + platform
+                if platform is not None
+                and platform != "native"
+                and platform != "default"
+                else ""
             )
-        ):
-            # Message for hard-supplied --build-cache message
-            if build_dir is not None:
-                gen_args = f" --build-cache {build_dir}"
-            else:
-                gen_args = " --ut" if self.build_type == BuildType.BUILD_TESTING else ""
-                gen_args += (
-                    " " + platform
-                    if platform is not None
-                    and platform != "native"
-                    and platform != "default"
-                    else ""
-                )
-            msg = f"'{self.build_dir}' is not a valid build cache. Generate this build cache with 'fprime-util generate{gen_args} ...'"
-            raise InvalidBuildCacheException(
-                msg,
-                self.build_dir,
-            )
+        msg = f"'{self.build_dir}' is not a valid build cache. Generate this build cache with 'fprime-util generate{gen_args} ...'"
+        raise InvalidBuildCacheException(
+            msg,
+            self.build_dir,
+        )
 
     def get_settings(
         self,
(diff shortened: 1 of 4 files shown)


src/fprime/fbuild/cmake.py
```diff
@@ -280,6 +280,8 @@ def generate_build(
             ),
             args.keys(),
         )
+        # Creating a file to mark the directory as a F Prime directory
+        open(build_dir / ".fprime-build-dir", "w")
```
thomas-bc: We should either close the file, or better imo is to use pathlib with for example 

```suggestion
        Path(build_dir / ".fprime-build-dir").touch()
```

src/fprime/fbuild/builder.py
```diff
@@ -112,7 +112,8 @@ def load(self, platform: str = None, build_dir: Path = None, skip_validation=Fal
         self.__setup_default(platform, build_dir)
         if not skip_validation and (
             not self.build_dir.exists()
-            or not (self.build_dir / "CMakeCache.txt").exists()
+            or (len(os.listdir(self.build_dir)) > 0 and
```
thomas-bc: What is the purpose of this `len(os.listdir)` check?

It makes it quite hard to read (not sure how the `or`/`and` combination is evaluated)
I wonder if we could skip this since we're checking for `.fprime-build-dir` below anyways

thomas-bc (changes requested):

thomas-bc:
@tylerrussin  Otherwise, this looks good to me – thanks!!

You'll need to run the code formatter on the files you changed, so that CI can pass.
```
pip install black
black ./path/to/file.py
```

src/fprime/fbuild/builder.py
tylerrussin: The `len(os.listdir)` checks if the directory is empty. In the event that you Ctrl+C after the directory has been made but before the .fprime-build-dir file has been created, you will end up with a situation where the purge command doesn't delete the directory

src/fprime/fbuild/cmake.py
```diff
@@ -280,6 +281,8 @@ def generate_build(
             ),
             args.keys(),
         )
+        # Creating a file to mark the directory as a F Prime directory
+        Path(build_dir / ".fprime-build-dir").touch()
```
thomas-bc: Looks like there are tests that show `build_dir` won't always be a Path object, so `build_dir / ".fprime-build-dir"` isn't a valid expression (can't use `/` on strings)

Should do instead:

```suggestion
        (Path(build_dir) / ".fprime-build-dir").touch()
```

src/fprime/fbuild/builder.py
```diff
@@ -110,27 +110,29 @@ def load(self, platform: str = None, build_dir: Path = None, skip_validation=Fal
...
-                    else ""
-                )
-            msg = f"'{self.build_dir}' is not a valid build cache. Generate this build cache with 'fprime-util generate{gen_args} ...'"
-            raise InvalidBuildCacheException(
-                msg,
-                self.build_dir,
+
+        if skip_validation:
+            return
+        if self.build_dir.exists() or (self.build_dir / ".fprime-build-dir").exists():
```
thomas-bc: This will evaluate to `True` and therefore not error if `self.build_dir` exists but `self.build_dir/.fprime-build-dir` doesn't, I'm not sure that this is what we want?

```suggestion
        if self.build_dir is not None and (self.build_dir / ".fprime-build-dir").exists():
```

thomas-bc (changes requested):

thomas-bc (approved):
LGTM! Thanks a lot 🚀 
Sorry for the delay on the review

PR #248, opened 2025-03-26
Allow `fprime-util new` to create components within deployments

### LINKED ISSUE #3321: Allow `fprime-util new` to create components within deployments
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**| `fprime-util new --component` |
---
## Feature Description

https://github.com/nasa/fprime-tools/blob/60587cc2eed3971006c00822569fd5cf346414f0/src/fprime/util/cookiecutter_wrapper.py#L91-L95

By checking only if the path is an fprime module, we are effectively restricting components from being created within a deployment. This should be allowed.

tylerrussin:
| | |
|:---|:---|
|**_Related Issue(s)_**| nasa/fprime #3321 |

---
## Change Description

This change allows the `fprime-util new --component` command to create new components within deployment directories. This change affects the 'prime-util new --deployment` command, but leaves it with the same functionality as before.

## Rationale

This change fixes nasa/fprime#3321

### DIFF
--- a/src/fprime/common/utils.py
+++ b/src/fprime/common/utils.py
@@ -1,4 +1,4 @@
-""" fprime.common.utils: defines common utility functions to be used across sub packages
+"""fprime.common.utils: defines common utility functions to be used across sub packages
 
 @author thomas-bc
 """
@@ -54,7 +54,7 @@ def check_path_is_within_fprime_module(path: Path, is_component: bool):
     statement = "register_fprime_"
     if is_component:
         statement += "module"
-        
+
     for line in lines:
         if line.startswith(statement):
             return True
--- a/src/fprime/common/utils.py
+++ b/src/fprime/common/utils.py
@@ -28,7 +28,7 @@ def replace_contents(filename, what, replacement, count=1):
         return new_file != changelog
 
 
-def check_path_is_within_fprime_module(path: Path, is_component: bool):
+def check_path_is_within_fprime_module(path: Path, is_component: bool = False):
     """Check if the current working directory is within an F prime component or deployment module
 
     This is done by checking if any line in the path/CMakeLists.txt file starts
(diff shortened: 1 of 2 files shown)


src/fprime/common/utils.py
```diff
@@ -28,16 +28,20 @@ def replace_contents(filename, what, replacement, count=1):
         return new_file != changelog
 
 
-def check_path_is_within_fprime_module(path: Path):
-    """Check if the current working directory is within an F prime module
+def check_path_is_within_fprime_module(path: Path, is_component: bool):
```
thomas-bc: Can we make this an optional variable, defaulting to False?

The benefit of doing so is that doesn't change the API for existing code that uses it, only adds functionality. It also suggests what users should select by default if they're not sure.

src/fprime/util/cookiecutter_wrapper.py
```diff
@@ -157,7 +159,10 @@ def new_component(build: Build, parsed_args: "argparse.Namespace"):
 def new_deployment(build: Build, parsed_args: "argparse.Namespace"):
     """Creates a new deployment using cookiecutter"""
 
-    if check_path_is_within_fprime_module(Path.cwd()) and not parsed_args.force:
+    if (
+        check_path_is_within_fprime_module(path=Path.cwd(), is_component=False)
```
thomas-bc: This code block doesn't need to change once we use the default value = False.

thomas-bc (commented):
LGTM! Thanks!

One very minor nit

src/fprime/common/utils.py
thomas-bc:
```suggestion
def check_path_is_within_fprime_module(path: Path, is_component: bool = False):
```

tylerrussin: Sounds good to me

thomas-bc (approved):

PR #259, opened 2025-05-24
Add --overwrite flag to impl command

### LINKED ISSUE #3187: Support overwriting stubs in fprime-util impl
| | |
|:---|:---|
|**_F´ Version_**| v3.5.1 |
|**_Affected Component_**| fprime-util |
---
## Feature Description

I propose adding support to `fprime-util impl` for overwriting the `ComponentName.cpp` and `ComponentName.hpp` files instead of creating the files as `ComponentName.template.cpp` and `ComponentName.template.hpp`.

This should be opt-in so that users don't inadvertently overwrite code they intend to keep. For example, gated behind `--overwrite`.

## Rationale

When I create a new component, I've gotten into the habit of customizing the `.fpp` file after using the cookiecutter and before generating the implementation.

```
# fprime-util new --component
[INFO] Cookiecutter source: using builtin
  [1/8] Component name (MyComponent): ComponentName
  [2/8] Component short description (Component for F Prime FSW framework.): 
  [3/8] Component namespace (NamespaceName): 
  [4/8] Select component kind
    1 - active
    2 - passive
    3 - queued
    Choose from [1/2/3] (1): 2
  [5/8] Enable Commands?
    1 - yes
    2 - no
    Choose from [1/2] (1): 2
  [6/8] Enable Telemetry?
    1 - yes
    2 - no
    Choose from [1/2] (1): 2
  [7/8] Enable Events?
    1 - yes
    2 - no
    Choose from [1/2] (1): 2
  [8/8] Enable Parameters?
    1 - yes
    2 - no
    Choose from [1/2] (1): 2
[INFO] Found CMake file at '[project]/Components/NamespaceName/CMakeLists.txt'
Add ComponentName to [project]/Components/NamespaceName/CMakeLists.txt at end of file? (yes/no) [yes]: 
Generate implementation files? (yes/no) [yes]: no
[INFO] Did not generate implementations for /workspaces/[project]/Components/NamespaceName/ComponentName. Please do so manually.
# [... manually edit ComponentName.fpp ...]
# fprime-util impl
[0/1] Re-running CMake...
[...]
ninja: no work to do.
Formatting [1/2] /workspaces/[project]/Components/NamespaceName/ComponentName/ComponentName.template.cpp
Formatting [2/2] /workspaces/[project]/Components/NamespaceName/ComponentName/ComponentName.template.hpp
```

But since `fprime-util impl` has different behavior than the implementation phase of `fprime-util new --component`, I now need to manually rename `ComponentName.template.cpp` to `ComponentName.cpp` and `ComponentName.template.hpp` to `ComponentName.hpp`.

This request is obviously very low priority... but it would sure be convenient!

tanuj-rai:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This PR adds --overwrite argument to fprime-util impl to rename .template.cpp and .template.hpp to .cpp and .hpp respectively. If renamed files exist, they are overwritten safely with user permission.

## Rationale
This change fixes https://github.com/nasa/fprime/issues/3187


src/fprime/fpp/impl.py
```diff
@@ -120,6 +120,14 @@ def fpp_generate_implementation(
             ],
         ),
     )
+    if args.overwrite:
+        for path in output_dir.glob("*.template.*"):
+            renamed = output_dir / path.name.replace(".template", "")
+            if renamed.exists():
+                print(f"[WARN] Skipping overwrite: {renamed}")
```
celskeggs: If I'm understanding correctly, this will prevent an existing `ComponentName.cpp` or `ComponentName.hpp` from being overwritten.

However, that's NOT what I asked for in https://github.com/nasa/fprime/issues/3187. I wanted an option that DOES overwrite these files and throws away all of the existing code in them. The `--overwrite` flag opts in to this behavior.

thomas-bc: Agreed. 

This feels like a somewhat dangerous thing to do. For example I like to hit "Arrow up" in my terminal to repeat commands, and I can already see myself repeating an `fprime-util impl --overwrite` by mistake. Good version control should prevent, but still.

I think a good UX could be to prompt users for confirmation instead of logging:

```
$ fprime-util impl --overwrite
[INFO] File contents in <insert filenames> will be overwritten. Do you wish to proceed? (yes/no): 
....
```

src/fprime/fpp/impl.py
```diff
@@ -203,4 +211,10 @@ def add_fpp_impl_parsers(
         help="Generate test helper code for hand-coding. Default to False, leveraging the test helpers autocoded by FPP.",
         required=False,
     )
+    impl_parser.add_argument(
+        "--overwrite",
+        action="store_true",
+        default=False,
+        help="Rename generated .template.cpp and .template.hpp files to .cpp and .hpp, if no conflicts",
```
thomas-bc:
```suggestion
        help="Overwrite contents of current CPP and HPP files. Use with caution.",
```

thomas-bc (changes requested):

thomas-bc:
You'll also need to run formatting on the files you modified for CI to pass. Instructions here https://github.com/nasa/fprime-tools?tab=readme-ov-file#black-formatter

src/fprime/fpp/impl.py
tanuj-rai: Thank you, @celskeggs for pointing out the mistake. If I understand it correctly, that --overwrite flag should fully replace existing ComponentName.cpp and ComponentName.hpp files, discarding any previous code. Just to confirm, would it work to first remove the existing files and then rename the `.template.*` files into place when --overwrite is used?.

thomas-bc: @Tanuj-rai https://docs.python.org/3/library/pathlib.html#pathlib.Path.replace
it looks like `Path.replace()` will take care of removing/overwriting if the file already exists. Just like a bash `mv` command would.

tanuj-rai: Thank you, @thomas-bc, for the guidance. I'm Sorry for the late response. I will try to fix the code as soon as possible.

tanuj-rai: Hi @celskeggs and @thomas-bc, I am really sorry for the significant delay. I’ve updated `impl.py` to use `Path.replace()` to overwrite `ComponentName.cpp` and `ComponentName.hpp` and added a “yes/no” prompt listing files to prevent accidental overwrites. Please let me know if this is the correct approach or if anything needs to be updated further.

thomas-bc:
The submitted code was not working - I went ahead and simplified it. No custom logging or prompt, users should use `--overwrite` with caution.

thomas-bc (approved):

celskeggs (approved):
Haven't tested this, but it looks good to me

PR #265, opened 2025-06-25
Load settings.ini environment in utility commands

### LINKED ISSUE #3081: fprime-util wrapped tools (fpp-*, visualize, etc) Do Not Get Passed Settings.ini Environment
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

The environment should be passed to all calls.  This should be done with the following steps:

1. Build a helper that can run wrapped utilities
2. Feed the helper settings.ini and environment.
3. Factor settings.ini one level higher than builder

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/3081

### DIFF
--- a/src/fprime/fpp/cli.py
+++ b/src/fprime/fpp/cli.py
@@ -9,6 +9,7 @@ import argparse
 from typing import Callable, Dict, List, Tuple
 
 from fprime.fpp.common import FppUtility
+from fprime.fbuild.builder import Build
 
 
 def run_fpp_check(
--- a/src/fprime/fbuild/gcovr.py
+++ b/src/fprime/fbuild/gcovr.py
@@ -4,6 +4,7 @@ import itertools
 import shutil
 import subprocess
 import sys
+import os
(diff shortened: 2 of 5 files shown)


src/fprime/fpp/cli.py
```diff
@@ -9,6 +9,7 @@
 from typing import Callable, Dict, List, Tuple
 
 from fprime.fpp.common import FppUtility
+from fprime.fbuild.builder import Build
```
LeStarch, line 12: Is this unused here, or was something missed?

LeStarch (approved):
One minor question

src/fprime/fpp/cli.py
thomas-bc: It's used for type hints and was doing the quoted version
```
(build: "Build")
```
which was not resolving correctly in VSCode

PR #268, opened 2025-07-03
Update/aggregation in fprime util

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This changes how Targets and contexts work in `fprime-util`.  Instead of presuming targets from a context path, this new strategy uses explicit enumerators operating on that path to determine what targets live there.  With some CMake metadata output, this results in the ability to build multiple modules registered in a single path.

This feeds into new `check` and `coverage` targets that no longer rely on `_check` targets to be provided by the build system.  

Behavior should fallback to original behavior for backwards compatibility.


src/fprime/fbuild/target.py
```diff
@@ -55,17 +61,23 @@ def is_supported(self, builder: "Build", context: Path):
...
+            builder, context_path
+        )
+        return self.any_supported(builder, enumerated_targets)
 
-    @abstractmethod
     def execute(
         self,
         builder: "Build",
-        context: Path,
+        context: TargetContext,
```
LeStarch: context_path: Path

src/fprime/util/cli.py
```diff
@@ -54,6 +54,7 @@ def utility_entry(args):
         print(f"[ERROR] Could not detect project directory for: {parsed.path}")
     except Exception as exc:
         print(f"[ERROR] {exc}", file=sys.stderr)
+        raise
```
LeStarch: Remove `raise`

src/fprime/util/cli.py
```diff
@@ -369,6 +370,7 @@ def parse_args(args):
             selected_parser[0].print_usage()
         else:
             selected_parser.print_usage()
+        raise
```
LeStarch: `raise`

LeStarch (commented):
Review comments.

kevin-f-ortega (approved):
LGTM. I'm liking the ability to run tests with the `-j` option.

PR #276, opened 2025-07-29
Fix Ninja buffered output

### LINKED ISSUE #3818: Ninja-based build buffers output for a long time
When building with Ninja (now the default), fprime-util buffers its output until the end of the build. This makes it appear as if the build system is hanging, when in fact it is making progress in the background. It would be better to report the progress to the console at shorter intervals.

For example, when I run
```
% fprime-util generate --ut -f
% fprime-util check --jobs 4
```
in the `fprime` directory, the system hangs during the compilation, then reports on the linking, then hangs while running the tests, then reports the test output. Better behavior would be to report progress during the compilation and testing.

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix https://github.com/nasa/fprime/issues/3818

Removes the old `_communicate()` method and replaces it with a call to subprocess.run() which either captures the output or prints it straight to console.

I am not sure why this complicated `_communicate()` method was there in the first place, so care should be taken in making sure it was not necessary for anything I may have missed? @LeStarch please confirm
The comments in the code lead me to believe it may have been necessary in the past to have CMake behave properly with output coloring and such, but it seems like this has been resolved. Maybe there also was a use case for needing to both capture the output and stream it to console at the same time, but this doesn't look to be the case anymore either.


LeStarch (commented):
I need to check usages a bit deeper, otherwise looks good!

LeStarch (approved):
This will work with the exception that "make" builds will not refresh cache appropriately.  We will merge as-is for `a2` and fix for `4.0`

PR #288, opened 2025-09-26
Skip ctest invocation if no tests are requested

### LINKED ISSUE #4012: Partial UT run is reported as an error
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|   |
---
## Problem Description

Using the following command:
```
cd fprime/Drv
fprime-util check
```

UTs in Drv pass, but the framework reports an error because not all UTs registered by the entire framwork have been ran

```
[...]
92/93 Test #92: Utils_ut_exe .....................................***Not Run   0.00 sec
      Start 93: Types_Circular_Buffer_ut_exe
Could not find executable /home/chammard/LENG-Work/fprime/fprime/build-fprime-automatic-native-ut/bin/Linux/Types_Circular_Buffer_ut_exe
93/93 Test #93: Types_Circular_Buffer_ut_exe .....................***Not Run   0.00 sec

2% tests passed, 91 tests failed out of 93


The following tests FAILED:
	  1 - Fw_Buffer_ut_exe (Not Run)
	  2 - Fw_Dp_ut_exe (Not Run)
[....]

	 92 - Utils_ut_exe (Not Run)
	 93 - Types_Circular_Buffer_ut_exe (Not Run)
Errors while running CTest
Output from these tests are in: /home/chammard/LENG-Work/fprime/fprime/build-fprime-automatic-native-ut/Testing/Temporary/LastTest.log
Use "--rerun-failed --output-on-failure" to re-run the failed cases verbosely.
[ERROR] Command '['ctest', '--test-dir', '/home/chammard/LENG-Work/fprime/fprime/build-fprime-automatic-native-ut', '--no-tests=error']' returned non-zero exit status 8.
```

It appears to be trying to run all UTs registered in the framework.

## Context / Environment

```
Operating System: Linux
CPU Architecture: x86_64
Platform: Linux-4.18.0-553.64.1.el8_10.x86_64-x86_64-with-glibc2.28
Python version: 3.9.20
CMake version: 3.26.0
Pip version: 25.0.1
Pip packages:
    fprime-tools==4.0.1
    fprime-gds==4.0.1
    fprime-fpp==3.0.0
Project submodules:
```

## Expected Behavior

CTest should either:
- only attempt to run the UTs that are registered within `Drv` if `--recursive` is given
- not run any UTs

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Fix nasa/fprime#4012

If no tests are requested (path does not contain tests and neither `--recursive` nor `--all` provided), skip ctest invocation and log the unexpected behavior

### DIFF
--- a/src/fprime/fbuild/check.py
+++ b/src/fprime/fbuild/check.py
@@ -70,6 +70,16 @@ class Check(EnumeratedAction):
             context and context != ["all"] and ".*" not in context
         ):
             cli_args.append("-V")
+
+        if not context:
+            # This only happens if the provided path does not contain tests
+            # and neither '--all' nor '--recursive' were provided
+            print(
+                "[INFO] No tests were found in the given context.",
+                "Did you mean to use `fprime-util check --recursive` or `--all` ?",
+            )
+            return
+
         # When not "all" append a regex to filter tests. .* works as a regex
         if context and context != ["all"]:
             test_regex = f"^({'|'.join(context)})$"
--- a/src/fprime/fbuild/check.py
+++ b/src/fprime/fbuild/check.py
@@ -9,7 +9,7 @@ import subprocess
 import sys
 from typing import Tuple, Dict, List
 
-from fprime.fbuild.types import NoSuchTargetException
+from fprime.fbuild.types import NoTargetFoundException
 from fprime.fbuild.target import (
     TargetContext,
     TargetScope,
@@ -55,7 +55,7 @@ class Check(EnumeratedAction):
         if not context:
             # This only happens if the provided path does not contain tests
             # and neither '--all' nor '--recursive' were provided
-            raise NoSuchTargetException(
+            raise NoTargetFoundException(
                 "No tests were found in the given context. Did you mean to "
                 "use `fprime-util check --recursive` or `--all` ?"
             )
--- a/src/fprime/fbuild/check.py
+++ b/src/fprime/fbuild/check.py
@@ -50,6 +50,16 @@ class Check(EnumeratedAction):
         args: Tuple[Dict[str, str], List[str], Dict[str, bool]],
     ):
         """Execute this target"""
+
+        if not context:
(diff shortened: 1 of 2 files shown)


src/fprime/fbuild/check.py
```diff
@@ -70,6 +70,16 @@ def execute_all(
...
             cli_args.append("-V")
+
+        if not context:
+            # This only happens if the provided path does not contain tests
+            # and neither '--all' nor '--recursive' were provided
+            print(
+                "[INFO] No tests were found in the given context.",
+                "Did you mean to use `fprime-util check --recursive` or `--all` ?",
+            )
+            return
```
thomas-bc: I have no opinion between this and `sys.exit(1)` instead. Thoughts?

LeStarch: This output should probably be raised as an Exception and thus would output and trigger an error at the top-level.  Running `check` and not performing tests seems like it should be an error case.  I believe we make CTest under the hood report error on no tests.

thomas-bc: used sys.exit() instead

src/fprime/fbuild/check.py
```diff
@@ -50,6 +50,16 @@ def execute_all(
         args: Tuple[Dict[str, str], List[str], Dict[str, bool]],
     ):
         """Execute this target"""
+
+        if not context:
+            # This only happens if the provided path does not contain tests
+            # and neither '--all' nor '--recursive' were provided
+            print(
```
LeStarch: I think this should be an exception, and let the top-level program perform the exit.  We've had issues before with these buried exit calls.

LeStarch (changes requested):

src/fprime/fbuild/check.py
thomas-bc: done

LeStarch (approved):

PR #305, opened 2026-04-10
Include cstdio in Deployment/Main.cpp cookiecutter for nasa/fprime#4987

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

Include cstdio in Deployment/Main.cpp cookiecutter for nasa/fprime#4987


bocchino:
Instead of including `cstdio` to support `printf`, should we switch to using `Fw::Logger::log`? This seems to be the better practice.

bocchino (approved):
Looks good!

LeStarch (approved):

PR #307, opened 2026-04-11
fprime-tools: deinit generated deployments during teardown

dcpagotto:
this updates the generated deployment template so it follows the expected shutdown lifecycle.

today generated topologies call tearDownComponents(state), but stop there. That means new deployments are created with an incomplete teardown path, missnig the deinitComponents(state) call.

this change adds deinitComponents(state) to the generated topology teardown, updates the header comment, and adds a regresison test to make sure generated code keeps the corret order.

I validated this locally on WSL, and the new test covers exactly this generated output case.

if anyone has diferent conetxt on the shutdown order here, please take a look, but this seems like the right fix to keep generated deployments aligned with the intended component lifecylce.

Thanks! Dhiego Pagotto...

### DIFF
--- a/test/fprime/fbuild/test_new_deployment.py
+++ b/test/fprime/fbuild/test_new_deployment.py
@@ -13,7 +13,6 @@ import unittest
 from pathlib import Path
 from unittest.mock import patch, MagicMock
 
-from cookiecutter.main import cookiecutter
 from fprime.fbuild.builder import Build
 from fprime.util.cookiecutter_wrapper import new_deployment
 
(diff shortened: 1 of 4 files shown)


LeStarch (dismissed):

test/fprime/fbuild/test_new_deployment.py
```diff
@@ -119,6 +120,36 @@ def test_new_deployment_in_subdir(self, mock_register, mock_cookiecutter):
             kwargs["extra_context"]["__include_path_prefix"], "Deployments/"
         )
 
+    def test_builtin_deployment_template_emits_shutdown_calls(self):
```
thomas-bc: need to either format the code with `pip install black` and `black path/to/file.py`, or remove this test. I'm not sure the test is adding any value here so I'd be in favor of removing it.

thomas-bc (changes requested):
Need to fix formatting and also add the same

test/fprime/fbuild/test_new_deployment.py
dcpagotto: Thank you!... you are right, I agree that the test brings more noise than value to this PR, so I will turn my focus back to the change... Thank you for the review... my regards.

dcpagotto:
Removed the extra test and kept the PR focused on the change.

thomas-bc (approved):

PR #310, opened 2026-05-12
Adding context path to enumerated context, specifying test dir correctly

LeStarch:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This makes execution of UTs faster by specifying a directory to enumerate in.

### DIFF
--- a/src/fprime/fbuild/target.py
+++ b/src/fprime/fbuild/target.py
@@ -13,10 +13,13 @@ from enum import Enum
 from pathlib import Path
 from typing import Dict, List, Set, Tuple
 
-from .enumerator import BuildTargetEnumerator
+from .enumerator import BuildTargetEnumerator, EnumeratedContext
 from .types import BuildType, NoSuchTargetException
 
-TargetContext = Tuple[List[str], Path]
+
+ExecuteContext = Path
+""" Context for execute calls (execute, is_supported) which represents the path where the target is being executed """
+
 
 
 class TargetScope(Enum):
@@ -47,7 +50,7 @@ class ExecutableAction(ABC):
         self.scope = scope
 
     @abstractmethod
-    def is_supported(self, builder: "Build", context_path: Path):
+    def is_supported(self, builder: "Build", context_path: ExecuteContext) -> bool:
         """Is this target supported via the given contextual path"""
         pass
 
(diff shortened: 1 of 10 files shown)


src/fprime/fbuild/target.py
```diff
@@ -358,10 +358,11 @@ class BuildSystemTarget(Target):
...
         self,
         builder: "Build",
-        context: TargetContext,
+        context_with_path: TargetContext,
         args: Tuple[Dict[str, str], List[str], Dict[str, bool]],
     ):
         # Ensure the cache is refreshed
+        context, context_path = context_with_path
         context = ["refresh_cache"] + context
         for build_target in context:
```
thomas-bc, line 372: bug here? should now be `context[0]`?

src/fprime/fbuild/target.py
```diff
@@ -16,7 +16,7 @@
 from .enumerator import BuildTargetEnumerator
 from .types import BuildType, NoSuchTargetException
 
-TargetContext = List[str]
+TargetContext = Tuple[List[str], Path]
```
thomas-bc: would be good to add a little docstring as to what this tuple represents

src/fprime/fbuild/target.py
thomas-bc, line 1: I think there's likely a bug in `any_supported()` now, it should also take the first argument of context 

https://github.com/lestarch/fprime-tools/blob/a9754f110fadee42074d5deb607c6fd0b5c42435/src/fprime/fbuild/target.py#L164

thomas-bc: if not, the type hint is wrong

src/fprime/fbuild/target.py
thomas-bc: Actually... I'd argue that a better approach would be to keep `TargetContext` as is, and introduce a new `TargetContextWithPath` tuple. Then the type hinting and logic becomes a lot clearer. 

Code like this worries me:

```
context, context_path = context
```

thomas-bc (changes requested):
I'm a bit worried that we're duct-taping a thing into a pre-existing type and the mental model is now hard to follow. I'd recommend, rather than modifying `TargetContext` to change its type entirely, introduce a new `TargetContextWithPath` tuple, and use that where needed.

src/fprime/fbuild/target.py
thomas-bc: nvm, it shouldn't, but has to do with the hard-to-follow mental model note above

src/fprime/fbuild/target.py
LeStarch: Why do is worry you?  Because context is used twice?

thomas-bc (approved):

PR #331, opened 2026-07-03
Make fprime-util format work for libraries

### LINKED ISSUE #4926: `fprime-util format` should work for libraries
| | |
|:---|:---|
|**_F´ Version_**| |
|**_Affected Component_**|  |
---
## Feature Description

An F´ library isn't a project, yet it still should be able to be formattable. The issue is that there's no easy way to go find the default F´ `.clang-format` file. 
Either libraries should provide their own, or maybe we could add in the behavior to go fetch that file and put it in the root of the library. Many possible solutions.

Expected behavior:
- clone an F´ library (such as https://github.com/fprime-community/fprime-zephyr) outside of a project
- run `cd fprime-zephyr && fprime-util format --dirs ./Svc` should work

Sammy-Dabbas:
Fixes nasa/fprime#4926.

## Problem

`fprime-util format` fails in an F Prime library (e.g. fprime-zephyr) for two stacked reasons:

1. `format` went through `load_build()`, which calls `Build.find_nearest_parent_project()`. A library has no root `CMakeLists.txt` with `project()`, so the command dies with "Could not detect project directory" before formatting ever runs.
2. The style file was hardcoded to `framework_path/.clang-format` from `settings.ini`, which libraries do not have.

## Fix

- Add `format` to `skip_build_loading()` (alongside `version-check`) so no build/project context is required, and remove it from the now-irrelevant `skip_build_cache_validation()`.
- New `locate_clang_format_file()` in `commands.py`: when a parent F Prime project is detected, use the framework's `.clang-format` exactly as before (no behavior change for deployments); otherwise walk up the directory tree from the working path for a `.clang-format`, mirroring clang-format's own discovery convention so libraries can carry their own style file.
- Guard `builder.settings` access in `ClangFormatter.execute()` so it tolerates `build=None`.

## Testing

- 4 new tests covering the library cases: `execute` with no build object, style-file discovery from inside a library, the missing-style-file error path, and an end-to-end `run_code_format` against a synthetic library with its own `.clang-format` (skipped if clang-format is unavailable).
- Formatter suite: 11 passed (7 existing + 4 new), re-verified after rebasing onto current `devel`.
- Manually exercised the CLI against synthetic libraries: format and `--check` behave correctly, a missing `.clang-format` produces a clear error, and the deployment/project case still resolves the framework style file (regression-checked).

Happy to adjust the discovery order or error text if you have a preferred convention.

### DIFF
--- a/src/fprime/util/commands.py
+++ b/src/fprime/util/commands.py
@@ -14,16 +14,13 @@ Current commands include:
 import argparse
 import sys
 from pathlib import Path
-from typing import Dict, List, Optional
+from typing import Dict, List
 import subprocess
 import platform
 import importlib.metadata
 
 
-from fprime.common.error import FprimeException
 from fprime.fbuild.builder import Build, InvalidBuildCacheException
-from fprime.fbuild.settings import IniSettings
-from fprime.fbuild.types import UnableToDetectProjectException
 from fprime.util.code_formatter import ClangFormatter
 from .versioning import VersionException, FPRIME_PIP_PACKAGES
 from fprime.util.cookiecutter_wrapper import (
@@ -157,38 +154,6 @@ def run_new(
     )
 
 
-def locate_clang_format_file(parsed: argparse.Namespace) -> Optional[Path]:
-    """Locate an explicit .clang-format style file, if the project declares one.
-
-    Inside an F´ project (which declares a `framework_path` in settings.ini) the
-    framework's `.clang-format` is used, preserving the historical behavior.
-    Inside a standalone F´ library there is no settings.ini, so this returns None
-    and lets clang-format discover the nearest `.clang-format` itself: format is
-    run with `--style=file`, which already searches each input file's parent
-    directories.
-
-    Args:
-        parsed: parsed input arguments
-
-    Returns:
-        Path to the framework's .clang-format when a project is detected, or None
-        to defer discovery to clang-format (standalone library).
-    """
-    try:
-        cmake_root = (
-            Path(parsed.root)
-            if parsed.root is not None
-            else Build.find_nearest_parent_project(Path.cwd())
-        )
-        settings = IniSettings.load(cmake_root / "settings.ini")
-        framework_path = settings.get("framework_path")
-        if framework_path is not None:
-            return Path(framework_path) / ".clang-format"
-    except (UnableToDetectProjectException, FprimeException):
-        pass  # Standalone library: let clang-format find the .clang-format itself
-    return None
-
-
 def run_code_format(
     build: Build,
     parsed: argparse.Namespace,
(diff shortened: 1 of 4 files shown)


src/fprime/util/commands.py
```diff
@@ -154,6 +157,51 @@ def run_new(
...
+            else Build.find_nearest_parent_project(Path.cwd())
+        )
+        settings = IniSettings.load(cmake_root / "settings.ini")
+        framework_path = settings.get("framework_path")
+        if framework_path is not None:
+            return Path(framework_path) / ".clang-format"
+    except (UnableToDetectProjectException, FprimeException):
+        pass  # Not in a project (e.g. a standalone library); fall back to search
+
+    # Library fallback: walk up from the working path searching for a .clang-format
```
thomas-bc: This is the default `clang-format` behavior so I think it'd be better to just delegate to it, and not rewrite the logic ourselves.

thomas-bc (changes requested):

src/fprime/util/commands.py
```diff
@@ -179,7 +227,7 @@ def run_code_format(
     }
     clang_formatter = ClangFormatter(
         "clang-format",
-        build.settings.get("framework_path", Path(".")) / ".clang-format",
+        locate_clang_format_file(parsed),
```
thomas-bc: would be useful to print out the found location in the case that `--verbose` is specified

Sammy-Dabbas:
Thanks. clang-format is already invoked with --style=file, so it discovers the nearest .clang-format from each input file's directory on its own. I dropped the manual walk up. The locate_clang_format_file now only resolves the framework's .clang-format when a project is detected and returns None for a standalone library, and the formatter skips its pre-flight existence check when there's no explicit framework file. The end-to-end library test still passes because clang-format finds the library's  config. A library shipping no .clang-format falls back to clang-format's default style rather than erroring, which matches its normal behavior.

thomas-bc (changes requested):
Looking at this in a bit more detail, I'm realizing that things have been off for a while (nothing to do with your PR - so thanks for unearthing this!). The code goes and look for `path/to/fprime/.clang-format`, but it's never actually used.

Since https://github.com/nasa/fprime-tools/pull/199, we only specify `--style=file` and never give an explicit path. So `fprime-util format` **always** delegates to clang-format to find the `.clang-format` file in a parent repo.

This is intended. Projects have their own `.clang-format` file, and libraries that want formatting should define their own as well. There's therefore no reason to try and find the `fprime/.clang-format` file. We should remove the logic to attempt to do it. Libraries can then run `fprime-util format` and it'll behave exactly the same: let clang-format try to find a `.clang-format` file in a parent directory.

Sammy-Dabbas:
Done in 5e99471. The --verbose location print is left out since we no longer
locate the file ourselves. There's nothing for us to print, and clang-format
resolves it internally. Happy to add it back in some form if you'd like.

Sammy-Dabbas:
Hi @thomas-bc, following up on this one. I pushed the changes you asked for; the framework .clang-format lookup is removed and discovery is delegated to clang-format via --style=file (5e99471, 6bac973). I left the --verbose found-location print out since the tool no longer resolves the file itself, but I am happy to add some form of it if you would like.

thomas-bc (approved):

PR #341, opened 2026-08-18
Add --existing to check --coverage: calculate coverage from existing data

lestarch-autobot:
| | |
|:---|:---|
|**_Related Issue(s)_**| n/a |
|**_Documentation Included (y/n)_**| y (option help strings) |

---
## Change Description

Adds an `--existing` switch to `fprime-util check --coverage` (per review feedback) that skips the test-run step and runs gcovr only, calculating coverage from `.gcda` data already on disk:

```python
class GcovrTarget(CompositeTarget):
    def option_args(self):
        return super().option_args() + [("--existing", "...no test re-run...")]

    def execute(self, builder, context, args):
        children = self.targets
        if args[2].get("--existing", False):
            children = [child for child in children if isinstance(child, Gcovr)]
        ...
```

Default behavior of `check --coverage` is unchanged.

## Rationale

Supports system-level (integration-test) coverage flows: one instrumented deployment run produces coverage data for all modules at once; the report split per module must then run gcovr repeatedly against the same on-disk data without re-running tests. Used by the F´ integration-coverage publisher (nasa/fprime-actions#43).

- `pytest test` passes (157 passed), including a new test asserting `--existing` skips the check step while plain `check --coverage` still runs it.
- Verified against a gcov-instrumented Ref deployment build: `fprime-util check --coverage --existing [--all] -r <deployment> --build-cache <deployment>/build-fprime-automatic-native` produces correct filtered reports from existing `.gcda` files, globally and per-module.

## Future Work

None.


thomas-bc (commented):
This is quite a niche use case (presumably for CI?) - would it be possible to have it be a sub-option of `fprime-util check --coverage --existing` or something? Otherwise users may get confused that `fprime-util coverage` doesn't do coverage

LeStarch (approved):

thomas-bc (approved):

PR #346, opened 2026-08-28
Remove dead code and improve status and error handling

thomas-bc:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |

---
## Change Description

This is the first PR in a set of AI-generated PRs aimed to improve quality and maintainability of this code base

## AI Usage

Devin identified and generated the content - I discussed and reviewed with it


sylvesterkaczmarek (commented):
`top_target=True` is still used by `Build` and covered by the existing global-target test, but this removes the only branch that leaves global CMake targets unprefixed. Those calls now resolve to `<module>_<target>` instead of `<target>`, making global targets unreachable. Please preserve that branch and keep the existing global-target regression test passing.

PR #347, opened 2026-08-29
Add `fprime-util sloc`: context-aware SLOC counting with FPP support

lestarch-autobot:
| | |
|:---|:---|
|**_Related Issue(s)_**|  |
|**_Documentation Included (y/n)_**| y (help text) |

---
## Change Description

Adds a new `fprime-util sloc` command that counts SLOC for F Prime projects, plus a Pygments lexer plugin for FPP/FPPI so pygount can count model files.

- **Module discovery**: a module is any directory whose `CMakeLists.txt` calls `register_fprime_*`, `add_library`, or `add_executable`; source files are attributed to the nearest enclosing module.
- **Context aware**:
  - component/module directory → counts just that module (no build cache needed)
  - deployment directory → counts the deployment and its linked dependency closure (build cache required; dependencies resolved from link files or the Ninja build graph)
  - project root → full segmentation into project / framework / `library_locations` from `settings.ini`, with per-section totals and a grand total (no build cache needed)
  - `--recursive` → counts all modules under the current directory (build cache required)
- **Classification** columns: `FPP`, `CMake`, `C/C++`, `C/C++ AC`, `C/C++ UT`, `C/C++ AC UT`, `C/C++ Total`, per-language `other (<lang>)` columns, `other (Total)`, and `Total`. UT code is identified via `test/`/`ut/` directories; autocoded (AC) files are counted from the build cache when available (shown as `-` otherwise).
- **Exclusions**: build caches, hidden dirs, virtualenvs, `__pycache__`, `node_modules`, `build-artifacts`, and git submodules (opt back in with `--include-submodules`).
- **Outputs**: CLI table, `--markdown-report`, `--json-report`, and `--excel-report` (`.xlsx`). `openpyxl` is an **optional** dependency (`pip install fprime-tools[excel]`); when missing, `--excel-report` reports a clear error instead of failing at import.
- `CMakeLists.txt` files are re-analyzed under a `.cmake` name so pygount counts them as CMake rather than plain text.

## Rationale

Projects need per-module/per-library SLOC accounting (including FPP models and autocoded output) for reporting and flight-software metrics; existing tools (cloc, pygount alone) do not understand F Prime module structure, `settings.ini` segmentation, or FPP.

- `pytest test/fprime/util/test_sloc.py` covers discovery, classification, UT detection, rollups, and all four renderers.
- Validated against the fprime repo (root, component, deployment scopes) and a stress-reference project with a generated build cache; results cross-checked against grep-based estimates and an FPP-adapted cloc and found in-family (differences explained by pygount treating punctuation-only lines as blank).
- Review focus: module discovery/attribution in `discover_modules`, deployment dependency closure (`ninja_link_tokens`), and AC classification.

## Future Work

None planned; report formats can be extended as needed.


thomas-bc (changes requested):
merge conflicts

src/fprime/fpp/lexer.py
```diff
@@ -0,0 +1,162 @@
...
+    Number,
+    Operator,
+    Punctuation,
+    String,
+    Text,
+    Whitespace,
+)
+
+FPP_KEYWORDS = (
+    "active",
```
thomas-bc, line 24: I'd rather avoid having to maintain this list of keywords here - is there a way to pull the grammar from https://github.com/fprime-community/fpp-tools

src/fprime/util/sloc.py
```diff
@@ -0,0 +1,879 @@
...
+import tempfile
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Callable, Dict, List, Optional, Tuple
+
+from pygount import SourceAnalysis, SourceState
+
+# openpyxl is optional: only needed for --excel-report
+try:
+    from openpyxl import Workbook
```
thomas-bc, line 33: sounds like a hassle to maintain and little use.. can we drop `--excel-report`? CSV could be enough.

thomas-bc (commented):
Looks neat - a few notes on maintainability. Maybe a solution would be to mark this as experimental ?

sylvesterkaczmarek (commented):
`count_autocoded()` returns `True` as soon as a build cache exists, even when `get_build_cache_path()` fails for individual discovered modules and those modules are silently skipped. The report then says autocode was counted and renders those modules' AC columns as `0` instead of unknown, which can undercount SLOC metrics; track availability per module (or mark/fail missing cache paths) and add a partial-cache regression.

PR #349, opened 2026-09-09
Validate cookiecutter names as FPP identifiers and harden CMake registration

### LINKED ISSUE #5874: fprime-util new: cookiecutter names are not validated as FPP identifiers and allow CMake injection
## Problem

`fprime-util new` pastes the name entered in the cookiecutter wizard verbatim into the build system:

```cmake
add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/<name>/")
```

`is_valid_name` in `src/fprime/util/cookiecutter_wrapper.py` only rejects a fixed list of characters. It does not block `;`, `(`, `)`, or newlines, so a name such as `x"); file(WRITE ...)` is accepted, written into `project.cmake` / `CMakeLists.txt`, and executed at configure time. Because these files are committed and shared, a wizard prompt becomes a way to plant arbitrary CMake in a repository.

Related problems in the same code path:

- The denylist accepts names that are not valid FPP identifiers (`1553Bus`, `Gps.v2`, `Système`), so generation succeeds and the failure surfaces later in `fpp-check` or the C++ compile with a confusing error.
- `..` segments and absolute paths are not rejected, so a bad relative path can register a directory outside the project.
- `add_to_cmake` writes `str(comp_path)`, which produces backslash-separated paths on Windows. `Comm\SwitchManager` is invalid CMake.
- The hook error message ("Do not use spaces or special characters") does not describe the rule that is actually applied.

## Proposed fix

- Require cookiecutter names to be FPP identifiers: letters, digits, and underscores, not starting with a digit. Keep the existing `is_valid_name` return contract (first offending character, or `"valid"`).
- Before writing to CMake, reject absolute paths, `.`/`..` parts, and the characters that can terminate or inject into a CMake quoted argument (backslash, double quote, dollar, semicolon, control characters). Existing parent directories with spaces or non-ASCII characters must still register normally.
- Write paths in POSIX form and treat the legacy backslash form as already registered so re-running the wizard on Windows does not append a duplicate line.
- Update the five `pre_gen_project.py` hook messages, the wizard prompts, and the `fprime-util new` help text to state the naming rule.
- Add unit tests covering injection attempts, accepted and rejected names, nested directories, duplicate detection, and error propagation.

## User-visible change

Names that start with a digit or contain dots were previously accepted and will be rejected. This is the correct rule for FPP identifiers and should be called out in the release notes.

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| Closes nasa/fprime#5874 |

---
## Change Description

- `is_valid_name` now requires cookiecutter names to be FPP identifiers: letters, digits, and underscores, not starting with a digit. It still returns the first offending character (or `"valid"`), so third-party templates that call it keep working.
- `add_to_cmake` validates the path before writing. It rejects absolute paths, `.`/`..` parts, and the characters that can terminate or inject into a CMake quoted argument (backslash, double quote, dollar, semicolon, control characters). Existing parent directories with spaces or non-ASCII characters still register normally.
- Registered paths are written in POSIX form. The backslash form written by older releases on Windows is recognised as already registered, so re-running the wizard does not append a duplicate line.
- `register_with_cmake` only catches the `relative_to()` error, so a CMake file outside the project root still fails the command with a nonzero exit as before.
- The five `pre_gen_project.py` hook messages, the wizard prompts in each `cookiecutter.json`, and the `fprime-util new` help text now state the naming rule.

## Rationale

Names entered in the `fprime-util new` wizard were pasted verbatim into `add_fprime_subdirectory("${CMAKE_CURRENT_LIST_DIR}/<name>/")`. The old denylist did not block `;`, parentheses, or newlines, so a name could inject CMake that runs at configure time and lands in committed project files. The denylist also accepted names that are not valid FPP identifiers (`1553Bus`, `Gps.v2`), which failed later in `fpp-check` or the C++ compile with a confusing error. Closes nasa/fprime#5874.

`test/fprime/util/test_cookiecutter_wrapper.py` grows from 3 to 46 tests covering injection attempts, accepted and rejected names, nested directories with spaces and non-ASCII characters, duplicate detection (including the legacy Windows form), and error propagation. Run with:

```
PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 PYTHONPATH=src python3 -m pytest test/fprime/util/test_cookiecutter_wrapper.py
```

Review focus:
- The unsafe-character set in `_UNSAFE_CMAKE_CHARS`. Only characters that are special inside a CMake quoted argument are rejected, so parent directories are not over-restricted.
- Manual check: run `fprime-util new --component` from a directory whose path contains a space and confirm it registers with the build system.

**User-visible change:** names that start with a digit or contain dots were previously accepted and are now rejected. This should be noted in the release notes.

## Future Work

None planned.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_019JF4eDGTZMTHXabs3oo8iU


thomas-bc:
This has never been flagged by users as a problem and is more code / regex that I'm comfortable reviewing for the priority that it is. This also speaks to things like validating Windows path –which we don't support– so I don't know that this was generated with our use case in mind. This is user input, so I don't know that CMake injection is an issue. Closing.

PR #352, opened 2026-09-10
Add fprime-util topology-check for static capacity limits

bitWarrior:
| | |
|:---|:---|
|**_Related Issue(s)_**| none |
|**_Documentation Included (y/n)_**| y (`fprime-util topology-check --help` entry in `help_text.py`) |

---
## Change Description

Adds a new `fprime-util topology-check` subcommand that statically checks a deployment's topology for capacity limits that today only surface as an `FW_ASSERT` when the flight software starts. It reads artifacts that `fprime-util generate`/`build` already leave in the build cache (the generated `<Deployment>TopologyAc.hpp/.cpp`, the topology dictionary JSON, and generated configuration headers) together with the FPP sources indexed by `locs.fpp`, then evaluates three rules:

| Rule | Level | What it checks |
|---|---|---|
| `health-ping-capacity` | error | Each `Svc.Health` instance's queue depth is >= its number of ping targets. Health is a queued component that only drains its queue on `Run`, so every ping reply of a cycle must fit. |
| `command-dispatcher-table` | error | `CMD_DISPATCHER_DISPATCH_TABLE_SIZE` is >= the number of commands in the dictionary. |
| `queued-component-burst` | warning (heuristic) | Each queued component instance's queue depth is >= the number of connections into its async input ports that assert on overflow. Ports declared with `drop` or `hook` are ignored. |

Output is one line per finding (`[OK]`, `[WARNING]`, `[ERROR]`) with a message that names the configuration knob to change. The exit code is 1 on any error, or on a warning when `--warnings-as-errors` is passed, so it drops straight into CI. Running from a directory with no generated topology, or with more than one, gives a clear error rather than a traceback.

Files:
- `src/fprime/fpp/topology_check.py`: parsers, model loader, rules, reporter, argparse registration.
- `src/fprime/util/cli.py`: registers the subcommand alongside the other fpp commands.
- `src/fprime/util/help_text.py`: long-form help entry.
- `test/fprime/fpp/test_topology_check.py`: 28 unit tests covering every parser, every rule's pass/fail/skip paths (including `drop`/`hook` exclusion, `$`-escaped instance names, nested modules), model loading from a fake build cache, and end-to-end exit codes.

## Rationale

Every F´ project that grows beyond the reference deployments eventually trips one of these limits, and the failure mode is unfriendly: the binary aborts a few seconds after launch with a bare assert line such as

```
Assert: ".../Svc/CmdDispatcher/CommandDispatcherImpl.cpp:68" 536956937
Assert: ".../Svc/Health/HealthComponentAc.cpp:898" 8
```

and the GDS simply never connects. The developer then has to map a file and line number back to an opcode table insert or a `Queue::FULL` status, work out which framework default was exceeded, and discover the `CONFIGURATION_OVERRIDES` mechanism to raise it. This PR came out of exactly that experience: a deployment built on the `CdhCore` subtopology crossed 150 commands and 25 ping targets after adding subsystem manager components, and both defaults asserted on the first run.

Why this benefits the whole community rather than one project:

- **Turns two of the most common "my deployment crashes at startup" questions into a one-line answer.** Both rules are exact static bounds, not guesses, and the message tells the user precisely what to change (`CdhCoreConfig.QueueSizes.health`, or `CommandDispatcherImplCfg.hpp` in the project config directory). New users hit these limits as soon as they scale past the tutorial, and experienced users hit them when subtopologies hide the defaults from view.
- **Makes framework defaults safe to keep small.** The framework can keep conservative table and queue sizes for memory-constrained targets, because projects that exceed them get told at build time instead of at run time.
- **CI-friendly by design.** A non-zero exit code and `--warnings-as-errors` let projects gate merges on it, catching regressions when someone adds a component or command without revisiting the config overrides.
- **Zero new dependencies.** It parses the generated code with regular expressions and reads `locs.fpp`, so it works with the native fpp binaries alone. `fpp-to-json` would be the ideal model source, but it requires a JVM that many developer machines and CI images do not have.
- **A home for future static checks.** Rules are plain functions collected in a `RULES` list, so further limits (parameter DB entries, telemetry packet sizes, file downlink buffer counts, and so on) can be added by contributors without touching the CLI plumbing.

- `pytest test/fprime/fpp/test_topology_check.py` (28 tests) and the full fpp test directory pass. Formatted with Black.
- Manually verified against two real deployments built on the `CdhCore`/`ComCcsds`/`DataProducts`/`FileHandling` subtopologies with 152 commands and 28 ping targets. With correct overrides it prints three `[OK]` lines; with the build-cache copies of the header and config temporarily set back to the defaults (25 and 150) it reports both errors and exits 1. Running from the project root, where two deployments exist, gives a clear error asking for a single deployment directory.
- Review focus: the regular expressions in `topology_check.py` are the coupling to the fpp-to-cpp output format (`namespace QueueSizes { enum {...} }`, `extern <Type> <name>;`, `X.set_P_OutputPort(i, Y.get_Q_InputPort(j))`, `X.setPingEntries(`). They match fpp v3.3.0 output; if the autocoder templates change, these are the lines to update. The third rule is documented as a heuristic and only produces a warning.
- One naming choice worth a look: the subcommand is `topology-check`, mirroring `fpp-check`. Happy to rename.

## Future Work

- Switch the model source to `fpp-to-json` via the existing `fprime.fpp.utils.fpp_to_json` layer once a JVM-free path exists, which would remove the dependence on generated-code formatting.
- Additional rules as the community reports other startup asserts with static bounds.
- Optionally chain the check after the `build` target for deployments via `CompositeTarget`, so it runs on every build. Left out of this PR to keep existing build semantics unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018JJ8BaNzZdsL7eNJcMbpfV


thomas-bc:
I think this check belongs in FPP and I actually believe there is a feature request for it out there. Closing this PR until the issue goes to CCB

